Issue #2, opened 2017-08-07
Re-organize this repo


evan2645:
- [x] 1. Top-level doc explaining what SPIFFE is. This document should link to all the following
- [x] 1. Core SPIFFE RFC
- [x] 1. X.509 SVID RFC
- [x] 1. SPIFFE Community Information
- [x] 1. svid-test

y2bishop2y:
How are we going to document the WL API? We moved all the docs to the NodeAgent. Should we just create a README.md that points to that proto buffer ?

evan2645:
@y2bishop2y I am thinking that we can create a similar standards doc which outlines those endpoints, their parameters, and their return format. I can link to the protobuf from there....

I realize that it duplicates the API signature to do this, however I think it's important, especially considering some implementors may not be familiar with/understand protobuf definition. I think IETF has got a standardized way to express these things... I will look into it

evan2645:
All done via #12 and #18

PR #4, opened 2017-08-08
import svid-test repo

drrt:


.gitignore
```diff
@@ -0,0 +1,11 @@
+test_results
```
evan2645: is this still valid? We should store test results in `svid-test/`

I think a lot of this .gitignore needs to be updated for that. Can .gitignore work in a subdirectory? If so, perhaps we can just move the whole thing

drrt: this .gitignore just lists filenames, which it will ignore no matter what directory they're in. .gitignore in subdirs is... controversial.

evan2645: Yea, the lack of scope is what concerns me, particularly since basically all of these apply only to a particular subdirectory. What are the drawbacks to putting .gitignore in `svid-test/`? If we can't do that, we should at least clean this up a little... for instance, I could see us putting *.log files elsewhere in this repository

evan2645 (approved):
Wooo, it's happeningggg

.gitignore
drrt: The main drawback is that's not where anyone ever expects to find it. Added directory names

evan2645:
Partially addresses #2

PR #18, opened 2017-08-18
Add/fix links in README and various standards docs

evan2645:
We can do this now that some of them are merged up


ajessup (approved):
LGTM

PR #19, opened 2017-08-18
Update GOVERNANCE.md

suniljames:
Reviewed the document and made a number of suggestions, particularly around TSC structure.

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -1,152 +1,104 @@
-The SPIFFE Project succeeds because of an open, inclusive and respectful community. Ideas and
-contributions are accepted according to their technical merit and alignment with project objectives,
-scope, and design principles.
-
-This document lists a non-exclusive set of guidelines to help ensure fairness and transparency in
-managing the SPIFFE Project.
+The SPIFFE Project succeeds because of an open, inclusive, and respectful community. Ideas and contributions are accepted based on their technical merit and alignment with project objectives, scope, and design principles. This document lists a non-exclusive set of guidelines to help ensure fairness and transparency in managing the SPIFFE Project.
 
 # Code of Conduct
 
-The SPIFFE community abides by the CNCF [code of conduct](/CODE-OF-CONDUCT.md). An excerpt:
+The SPIFFE community abides by the Cloud Native Computing Foundation's [code of conduct](/CODE-OF-CONDUCT.md). An excerpt follows:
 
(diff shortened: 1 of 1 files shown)


GOVERNANCE.md
```diff
@@ -1,152 +1,104 @@
...
 
-TSC size nor terms are limited. Committee members are added (or removed ) by the consensus of the
-existing TSC members. TSC members may remove themselves voluntarily at any time.
-
-No more than 2 TSC members may be affiliated with the same employer.
+The TSC adheres to the following:
 
+* The TSC is comprised of at least five (5) members.
+* No more than 2 TSC members may be affiliated with the same organization.
+* At least 40% of the TSC must be represented by organizations who currently have a SPIFFE reference implementation deployed in production.
```
ajessup: In principle I like this, but technically won't this make it impossible to form a TSC for the next several months?

GOVERNANCE.md
```diff
@@ -1,152 +1,104 @@
...
+  * Is consistent with the goals and direction of the Project. This requires not just code and architectural correctness, but also ensuring that it does not introduce "scope creep," and does not unduly affect existing users.
+* Once a PR has the requisite approvals, the last approving maintainer is responsible for merging the change (however, the PR's author must ensure the change is merge-ready).
+
+Maintainers are documented in each repository's [CODEOWNERS](https://help.github.com/articles/about-codeowners/) file. To become a maintainer, one must:
 
 * Work in a helpful and collaborative way with the community
-* Have a track record of providing constructive feedback on other's PRs
-* Have submitted a minimum of 50 PRs themselves
+* Have a track record of providing constructive feedback on others' PRs
+* Have submitted at least 50 PRs themselves
```
evan2645: 50 feels like a lot, I might take that down to 20 or so, but whatevs we can always adjust it later

GOVERNANCE.md
evan2645: suggest `SPIFFE reference implementation` -> `SPIFFE implementation`

evan2645 (approved):
lgtm

GOVERNANCE.md
suniljames: Yes, we have a bootstrapping issue. Perhaps these rules go into place upon the first election process in nine months?

drrt: same

PR #27, opened 2017-08-23
In the X.509 SVID spec, clarify key usage requirements for CAs.

briansmith:
Signing certificates don't need to set cRLSign, and the probably shouldn't if
they know they aren't going to issue CRLs.

Leaf certificates don't need to set `keyAgreement` or `keyEncipherment` if they
are never going to be used for key agreement or key encipherment. For example,
a leaf certificate with an ECC key is probably only going to be used for signing
ECDSA/EdDSA signatures. A leaf certificate with an RSA key likely should have
`keyEncipherment` so that it can be used for RSA key exchange, but if only
(EC)DHE-based key exchange is to be used then it isn't needed.

### DIFF
--- a/standards/X509-SVID.md
+++ b/standards/X509-SVID.md
@@ -70,9 +70,9 @@ The key usage extension defines the purpose of the key contained in the certific
 
 The key usage extension MUST be set on all SVIDs, and MUST be marked critical.
 
-SVID signing certificates MUST set `keyCertSign` and `cRLSign`. They MUST NOT set `keyEncipherment` or `keyAgreement`. This helps ensure that they cannot be used for authentication purposes.
+SVID signing certificates MUST set `keyCertSign`. They MAY set `cRLSign`.
 
-Leaf SVIDs MUST set `keyEncipherment`, `keyAgreement`, and `digitalSignature`. They MUST NOT set `keyCertSign` or `cRLSign`.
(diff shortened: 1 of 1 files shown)


rra:
> A leaf certificate with an RSA key likely should have
> keyEncipherment so that it can be used for RSA key exchange, but if only
> (EC)DHE-based key exchange is to be used then it isn't needed.

Maybe say that in the proposed wording?  Otherwise, this is a bit of a land mine for folks who haven't spent a lot of time reading the standard, since a lot of implementations don't bother to check these constraints (but then some do).  So it's easy to accidentally create certs that can be used by some implementations but not others.

evan2645 (approved):
Agreed on @rra's comment, though I don't think it blocks. In general, I think we could do a better job of explaining things like this.

Thanks for the contrib. This change needs at least one more approver.

dudedad:
Agreed on limiting MUST requirements and allowing for differing use cases

briansmith:
I updated the PR based on the feedback. Basically I tried to minimize the amount of unnecessary requirements being imposed. In particular, even though it probably doesn't make sense for a SVID signing certificate to have keyAgreement, etc., capabilities, it also doesn't hurt anything as far as SPIFFE is concerned.

mattmoyer (approved):
Looks good to me (discussed in SIG-Spec 2017-10-04).

PR #29, opened 2017-08-23
In the X.509 SVID spec, make EKU requirements more flexible.

briansmith:
Leave the door open for future work on EKU restrictions in intermediate
CA certifictes.

An empty EKU field means "all EKUs are allowed." Therefore, it doesn't
make much sense to allow an empty EKU but then forbid EKUs that have
EKUs other than id-kp-serverAuth and id-kp-clientAuth; if other EKUs
are truly harmful to be included then a missing EKU extension should be
forbidden too.

### DIFF
--- a/standards/X509-SVID.md
+++ b/standards/X509-SVID.md
@@ -77,9 +77,10 @@ Leaf SVIDs MUST set `keyEncipherment`, `keyAgreement`, and `digitalSignature`. T
 ### 4.4. Extended Key Usage
 This extension indicates one or more purposes for which the key contained in the certificate may be used, in addition to or in place of the basic purposes indicated in the key usage extension. It is defined in [RFC 5280, section 4.2.1.2][6].
 
-The extended key usage extension applies to leaf certificates only. Leaf SVIDs SHOULD include this extension, and it MAY be marked as critical. When included, fields `id-kp-serverAuth` and `id-kp-clientAuth` MUST be set, and all other fields MUST NOT be set.
+Leaf SVIDs SHOULD include this extension, and it MAY be marked as critical. When included, fields `id-kp-serverAuth` and `id-kp-clientAuth` MUST be set.
 
-Signing certificates MUST NOT include an Extended Key Usage extension.
+Signing certificates SHOULD NOT include an Extended Key Usage extension because the processing of the Extended Key Usage extension
+in intermediate CA certificates varies widely across X.509 certificate validation libraries.
 
 ## 5. Validation
 This section describes how an X.509 SVID is validated. The procedure uses standard X.509 validation, in addition to a small set of SPIFFE-specific validation steps.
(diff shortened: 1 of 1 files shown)


rra:
This looks reasonable to me, although do you have some specifics on the statement here that EKU restrictions vary widely on intermediate certs?

briansmith:
> do you have some specifics on the statement here that EKU restrictions vary widely on intermediate certs?

In Microsoft's implementation of X.509, and in Firefox's (mozilla::pkix) and the Rust [webpki](https://github.com/briansmith/webpki) implementation, any EKU constraints in intermediate CA certificates are enforced on the certificates below. So, for example, if the intermediate CA has only id-kp-serverAuth and the end-entity certificate has both id-kp-serverAuth and id-kp-clientAuth, then they will only trust the certificate for id-kp-serverAuth. However, this isn't universally implemented. I believe that OpenSSL, for example, does things differently.

rra:
Thank you!  I think that would be a useful comment to add to the spec itself in some abbreviated form, like: "Some X.509 implementations will impose EKU constraints in intermediate CA certificates on all certificates below them in the trust chain, and others will not."

standards/X509-SVID.md
```diff
@@ -77,9 +77,10 @@ Leaf SVIDs MUST set `keyEncipherment`, `keyAgreement`, and `digitalSignature`. T
 ### 4.4. Extended Key Usage
 This extension indicates one or more purposes for which the key contained in the certificate may be used, in addition to or in place of the basic purposes indicated in the key usage extension. It is defined in [RFC 5280, section 4.2.1.2][6].
 
-The extended key usage extension applies to leaf certificates only. Leaf SVIDs SHOULD include this extension, and it MAY be marked as critical. When included, fields `id-kp-serverAuth` and `id-kp-clientAuth` MUST be set, and all other fields MUST NOT be set.
+Leaf SVIDs SHOULD include this extension, and it MAY be marked as critical. When included, fields `id-kp-serverAuth` and `id-kp-clientAuth` MUST be set.
 
-Signing certificates MUST NOT include an Extended Key Usage extension.
+Signing certificates SHOULD NOT include an Extended Key Usage extension because the processing of the Extended Key Usage extension
+in intermediate CA certificates varies widely across X.509 certificate validation libraries.
```
evan2645: Erroneous newline?

evan2645 (commented):
Agreed with @rra, a "for instance" example would be useful here since we are giving a justification

briansmith:
I removed the extra newline and added the suggested statement.

evan2645 (approved):
LGTM, needs one more approver

standards/X509-SVID.md
```diff
@@ -77,9 +77,9 @@ Leaf SVIDs MUST set `keyEncipherment`, `keyAgreement`, and `digitalSignature`. T
 ### 4.4. Extended Key Usage
 This extension indicates one or more purposes for which the key contained in the certificate may be used, in addition to or in place of the basic purposes indicated in the key usage extension. It is defined in [RFC 5280, section 4.2.1.2][6].
```
justinburke, line 78: I think this should be section 4.2.1.12.

justinburke:
LGTM.

PR #30, opened 2017-08-23
In the X.509 SVID spec, improve basic constraints requirements.

briansmith:
Don't require a leaf certificate to have a basic constraints extension
given that the default, without an extension, is CA=false. Remove
"for authentication purposes" because it seems to be redundant. (If it
isn't redundant than it is unclear what other purpose a leaf would be
validated for.)

### DIFF
--- a/standards/X509-SVID.md
+++ b/standards/X509-SVID.md
@@ -56,7 +56,7 @@ Leaf and signing certificates carry different X.509 properties - some for securi
 ### 4.1. Basic Constraints
 The basic constraints X.509 extension identifies whether the certificate is a signing certificate, as well as the maximum depth of valid certification paths that include this certificate. It is defined in [RFC 5280, section 4.2.1.9][3].
 
-Valid X.509 SVIDs (both leaf and signing certificates) MUST NOT set the `pathLenConstraint` field. Signing certificates MUST set the `CA` field to `true`, and leaf certificates MUST set the `CA` field to `false`.
+SVID signing certificates MUST set the `CA` field to `true` and MUST NOT set the `pathLenConstraint` field. Leaf certificates MUST either omit the basic constraints extension or have an empty basic constraints field (an empty basic constraints field is the only correct encoding of a basic constraint with `CA` set to false.)
 
 ### 4.2. Name Constraints
 Name constraints indicate a namespace within which all SPIFFE IDs in subsequent certificates in a certification path MUST be located. They are used to limit the blast radius of a compromised signing certificate to the named trust domain(s), and are defined in [RFC 5280, section 4.2.1.10][4]. This section applies to signing certificates only.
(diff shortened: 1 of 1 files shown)


standards/X509-SVID.md
```diff
@@ -56,7 +56,7 @@ Leaf and signing certificates carry different X.509 properties - some for securi
 ### 4.1. Basic Constraints
 The basic constraints X.509 extension identifies whether the certificate is a signing certificate, as well as the maximum depth of valid certification paths that include this certificate. It is defined in [RFC 5280, section 4.2.1.9][3].
 
-Valid X.509 SVIDs (both leaf and signing certificates) MUST NOT set the `pathLenConstraint` field. Signing certificates MUST set the `CA` field to `true`, and leaf certificates MUST set the `CA` field to `false`.
+SVID signing certificates MUST set the `CA` field to `true` and MUST NOT set the `pathLenConstraint` field. Leaf certificates MUST either omit the basic constraints extension or have an empty basic constraints field (an empty basic constraints field is the only correct encoding of a basic constraint with `CA` set to false.)
```
rra, line 59: [citation needed].  RFC 5280 says:

> This extension MAY appear as a critical or non-critical extension in end entity certificates.

and allows setting it to false.  What's the reference for it being more correct to omit the basic constraints entirely rather than being explicit?

rra: Oh, never mind, I see -- this is a property of DER.  I was looking at too high of a protocol level.  Yes, this looks good to me.

briansmith: @rra Right, the DER rules make it an error to explicitly encode CA = false. See https://github.com/awslabs/certlint/issues/11 for more context about why I suggested rewording things.

standards/X509-SVID.md
```diff
@@ -92,7 +92,8 @@ In order to perform path validation, it is necessary to possess the public porti
 ### 5.2. Leaf Validation
 When authenticating a resource or caller, it is necessary to perform validation beyond what is covered by the X.509 standard. Namely, we must ensure that 1) the certificate is a leaf certificate, and 2) that the signing authority was authorized to issue it.
 
-When validating an X.509 SVID for authentication purposes, the validator MUST ensure that the `CA` field in the basic constraints extension is set to `false`, and that `keyCertSign` and `cRLSign` are not set in the key usage extension. The validator must also ensure that the scheme of the SPIFFE ID is set to `spiffe://`.
+When validating an X.509 SVID (leaf), the validator MUST ensure that the X509 version is 3 (encoded as 0x02). The validator MUST ensure
+that the `CA` field in the basic constraints extension, if present, is NOT set to `true`. The validator must also ensure that the scheme of the SPIFFE ID is set to `spiffe://`.
```
evan2645: Erroneous line break? The current doc doesn't constrain line length, though we can add breaks throughout if there is precedent for doing so

evan2645 (approved):
Awesome, thanks for the contrib!

This change needs at least one more approver.

briansmith:
I removed the extra newline that was added.

evan2645 (approved):

standards/X509-SVID.md
dudedad: LGTM.

justinburke (approved):

Issue #26, opened 2017-08-23
Clarify cross-signing of SVID signing certificates


briansmith:
The draft spec says "the SPIFFE ID of the signing certificate MUST be equal to the leaf certificate’s SPIFFE trust domain". However it also says "A signing certificate MAY be used to issue further signing certificates in the same or different trust domains."

Let's consider two trust domains, `spiffe://a` and `spiffe://b`. `spiffe://b` wishes to issue a certificate for a service `spiffe://a/some-service`, but it can't because it's SPIFFE ID isn't the same trust domain. However, it seems like it can easily circumvent that restriction by first issuing its own "spiffe://a" sub-CA certificate and then issuing the `spiffe://a/some-service` certificate from that sub-CA certificate.

Either the specification should clarify things to explain why this isn't possible, or why it is intentionally allowed and why it is not dangerous. Or, if this scenerio isn't intentionally allowed then maybe some additional requirements on validation are required.

evan2645:
This was not intentional, and this certainly seems to be a problem. A proposed solution from @briansmith is to pin CA bundles to trust domains. When validating a leaf SVID, ensure that the "correct" bundle is loaded by matching it with the trust domain in the leaf.

Will probably need to put some thought into how this impacts workload API etc.

evan2645:
This has long been corrected, and SVID validation no longer relies on matching anything up the chain. The recommended validation semantics have been in place for years now. Closing this out as stale. Thank you for having pointed this out so early on @briansmith

PR #42, opened 2017-10-05
Remove SPIFFE ID matching & SKI/AKI matching from X.509 spec.

briansmith:
Matching the SPIFFE ID and SKI/AKI in the leaf to its immediate
issuing certificate is non-standard (in X.509 terms) behavior that is
difficult to implement correctly and securely, especially given the
wide variety of certificate validation libraries in use. Other
mechanisms will be needed to achieve the goals these checks were
intended for. Although those new mechanisms haven't been finished yet,
that doesn't need to block the removal of these non-working checks now.

This is one step towards resolving #26.


evan2645 (approved):
LGTM. Probably better that we don't have folks performing this validation when we know we're going to change it anyways.

This change needs one additional approver

mattmoyer:
+1

PR #46, opened 2017-11-04
Mark workload API as a WIP

evan2645:
WISOTT

Don't want to send the wrong message. Workload API exists, a formal spec is pending


ajessup (approved):
LGTM

PR #51, opened 2017-11-30
Update SPIFFE-ID.md

kunzimariano:
While I was trying to point to one specific section of this doc I noticed that the links in the TOC are broken. Dots (.) won't work when trying to generate an anchor link.

For example:
- Doesn't work
https://github.com/spiffe/spiffe/blob/master/standards/SPIFFE-ID.md#2.1.-trust-domain
- Works
https://github.com/spiffe/spiffe/blob/master/standards/SPIFFE-ID.md#21-trust-domain

### DIFF
--- a/standards/SPIFFE-ID.md
+++ b/standards/SPIFFE-ID.md
@@ -11,15 +11,15 @@ This document, in particular, serves as the core specification for the SPIFFE st
 For more general information about SPIFFE, please see the [Secure Production Infrastructure Framework for Everyone (SPIFFE)](SPIFFE.md) standard.
 
 ## Table of Contents
-1\. [Introduction](#1.-introduction)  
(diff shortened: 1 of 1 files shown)


evan2645 (approved):
thanks!! here's to hoping we don't melt when we get to double digits :)

evan2645:
Oh... one minor thing. This change removes the spaces at the end of the ToC lines, but those are required to keep each item on a new line - otherwise they run together

kunzimariano:
Ooops let me correct that.

kunzimariano:
Should be OK now. Feel free to squash it :)

evan2645 (approved):
Merging this w/o an additional approval since there are no material changes to the spec itself

Issue #25, opened 2017-08-23
Clarify why the SVID signing certs "MUST NOT set the `pathLenConstraint` field"


briansmith:
In particular, `pathLenConstraint` set to 0 would mean that the SVID signing cert is valid for signing leaf certs only and can't be used to cross-sign other SVID signing certs (for other trust domains). This seems like a valid restriction to want to place on a SVID signing certificate so either (a) `pathLenConstraint` should be allowed, or (b) the spec should explain *why* path length constraints aren't allowed.

evan2645:
I had wondered about this too when writing the doc. That requirement was copy-pasted from the original SVID spec. Looking at Git history, that language came in w/ the initial commit and was sourced from an unlinked Google doc. Perhaps @ajessup can shed some light on the reasoning here?

There is a use case in which SPIFFE signing authorities may be chained together, and I suspect that this language came from a desire to support that use case, though it does seem fairly heavy handed. As an operator, I would like to set `pathLenConstraint` to `0` on signing authorities that I know I won't be chaining, or if chaining is not part of my SPIFFE implementation/deployment pattern.

Preskton:
I'm new to SPIFFE, but interested in this as well, especially as it relates to integrated into an existing PKI hierarchy.  Since signing cert must be issued with a "CA" type, it seems this would preclude someone from using a PKI where higher order CAs apply a path length constraint.

evan2645:
@Preskton yes it would be problematic in the event that the user would like higher order CAs to be SPIFFE-compliant (this is not a requirement at the moment).

Since some time has gone by and we still haven't heard any good reasons for keeping it, I propose we strike the requirement from the spec. Would you like to raise a PR where we can discuss this further?

evan2645:
#52 has been merged, which specifies pathLenConstraint as optional. Marking this issue as resolved. Thanks very much @briansmith and @Preskton!

PR #52, opened 2017-12-12
Remove restriction on `pathLenConstraint`

Preskton:
Hi, all - following the discussion in #25, raising a PR to remove the `pathLenConstraint` restriction on certs. If I was going to write a user story, it'd look something like this:

As an enterprise IT security manager with an existing PKI, I would like SPIFFE keys to allow the inclusion of `pathLenConstraint`s so that I can issue a SPIFFE signing certificate based on a private key infrastructure that has marked this constraint as `critical`. This would allow me to use my existing root and other infrastructure as the anchor for issuing SPIFFE CAs.

### Acceptance Criteria

- SPIFFE certificates that include the `pathLenConstraint` that otherwise meet the criteria to be SPIFFE compliant should be accepted as valid SPIFFE certificates.

### DIFF
--- a/standards/X509-SVID.md
+++ b/standards/X509-SVID.md
@@ -56,7 +56,7 @@ Leaf and signing certificates carry different X.509 properties - some for securi
 ### 4.1. Basic Constraints
 The basic constraints X.509 extension identifies whether the certificate is a signing certificate, as well as the maximum depth of valid certification paths that include this certificate. It is defined in [RFC 5280, section 4.2.1.9][3].
 
-Valid X.509 SVIDs (both leaf and signing certificates) MUST NOT set the `pathLenConstraint` field. Signing certificates MUST set the `CA` field to `true`, and leaf certificates MUST set the `CA` field to `false`.
+Valid X.509 SVIDs (both leaf and signing certificates) MAY set the `pathLenConstraint` field. Signing certificates MUST set the `CA` field to `true`, and leaf certificates MUST set the `CA` field to `false`.
 
 ### 4.2. Name Constraints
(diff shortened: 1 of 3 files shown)


Preskton:
CLA'd!

standards/X509-SVID.md
```diff
@@ -104,7 +104,7 @@ Extension | Field | Description
 ----------|-------|------------
 Subject Alternate Name | uniformResourceIdentifier | This field is set equal to the SPIFFE ID. Only one instance of this field is permitted.
 Basic Constraints | CA | This field must be set to `true` if and only if the SVID is a signing certificate.
-Basic Constraints | pathLenConstraint | This field must not be set.
+Basic Constraints | pathLenConstraint | This field may be be set if the implementor wishes to enforce a finite CA hierarchy depth, for example, if inherited from an existing private key infrastructure (PKI).
```
evan2645, line 107: My impression is that `pathLenConstraint` is not required to be set on subsequent "downstream" intermediates to which it applies. To prevent confusion, perhaps we can simply omit the example?

y2bishop2y (approved):

standards/X509-SVID.md
simo5: Maybe add a warning on what is the problem when setting it, so people are warned.

y2bishop2y:
The reasoning for this was that if intermediates would be added, the constraint values would have to be updated up and down the chain. IE if you had a pathlenConstraint of 2 and added an intermediate  all the values would have to be updated down the chain.

justinburke (approved):

justincormack:
LGTM

mcpherrinm:
I think pathLenConstraint makes sense in some cases but not others. My org's pki doesn't set it on offline roots, but we do set it on all online CAs. If we wanted to change our pki hierarchy, we'd do it from an offline root, but we don't want online attackers to be able to mint a CA to sign further certs.

I don't think SPIFFE should dictate how our PKI is run. As such, I don't think there's a 1-size-fits-all policy that can work here, and this needs to be optional.

Issue #56, opened 2018-01-22
Clarify SAN usage in the X509-SVID specification


az82:
The X509-SVID specification mandates that a subject alternative name (SAN) is provided and makes no mention of the subject DN.

RFC 5280, Section 4.2.1.6, §3 mandates that if the subject DN is empty, the SAN must be marked as critical (See https://tools.ietf.org/html/rfc5280#section-4.2.1.6 for reference).

Not marking the SAN as critical leads validation errors in conforming implementations, for instance in Java with the default Sun* providers or BouncyCastle.

Please clarify:
- whether the subject DN is optional
- that if the subject DN is not set, the SAN must be marked as critical

ajessup:
Agree, the spec is ambiguous here. As per the conversation in Slack, we should probably update the spec here to note that:

* The Subject DN is optional
* If the Subject DN is not set, then the SAN MUST be marked as critical

justinburke:
When there is no Subject DN, I agree that the SVID should be updated to mark the SAN extension as critical (ie. required). Doing so will bring us into compliance with RFC 5280.

I don't see any discussion of the Subject field in SVID doc and I haven't formed an opinion yet on whether the subject DN should be optional. I need to do a bit of research. Comments/insights welcome.

I'll note that Evan mentioned in Slack that Subject CN's are still in use by folks today. But the CN field is deprecated. In light of that, we could discourage its use in SVID or out right ban it.

az82:
Both subject and the CN field are widely used in the enterprise. I know of several large PKIs at our customers that use both, even though CN is deprecated according to https://tools.ietf.org/html/rfc2818. Usually such PKI subjects reference LDAP nodes and are hard to change.
I should also point out that the reason CN is deprecated is that it was used for matching host names, a practice no longer supported by Chrome and Firefox.

I recommend not imposing any more restrictions on the SVIDs than strictly necessary. Being lenient makes using SPIFFE with existing PKIs easier.

ajessup:
cc // @Preskton for feedback

Preskton:
Like the idea of SAN as critical if subject DN isn't set.

PR #57, opened 2018-01-24
Updates X509-SVID spec to note when SAN field should be critical

ajessup:
Addresses #56

### DIFF
--- a/standards/X509-SVID.md
+++ b/standards/X509-SVID.md
@@ -102,7 +102,7 @@ This document set forth conventions and standards for the issuance and validatio
 ## Appendix A. X.509 Field Reference
 Extension | Field | Description
 ----------|-------|------------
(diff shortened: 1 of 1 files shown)


justinburke:
Looks like this PR only partially addresses #56. That issue is asking for clarification on when the SAN field should be marked as critical as well as whether a Subject field is needed. And even with this PR, there's no discussion of the Subject field.

How about we hold off on this PR until we've reached a consensus on both points (to be discussed in comments of #56)?

ajessup:
This seems fair to me. As you note, this PR expressly avoids specifying if a Subject field may be supplied at all. If we can provide such guidance, this PR becomes simpler.

evan2645:
This is now being addressed in #75 

Please see those changes and leave comments there if you have any. thanks!

Issue #62, opened 2018-02-11
IANA registration for spiffe


mhausenblas:
It's a good practice—if not a requirement considered by some—to officially register a URI scheme, see also [BCP35](https://tools.ietf.org/html/bcp35) for guidance.

Since SPIFFE claims the `spiffe` URI scheme, we should kick off the [IANA registration](https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml) for this scheme.

suniljames:
@mhausenblas do you think it suffices if we e-mail IANA a link to an updated version of [this SPIFFE page](https://github.com/spiffe/spiffe/blob/master/standards/SPIFFE-ID.md) that include relevant bits as exemplified in the [sgn provisional application](https://www.iana.org/assignments/uri-schemes/prov/sgn)? Process looks to be fairly lightweight.

mhausenblas:
Yes @suniljames, if as per [BCP35, sect 4](https://tools.ietf.org/html/bcp35#section-4) the requirements are fulfilled the rest is really just a couple of mails back and forth. I suggest to think a little bit about who will be the contact; it should be someone trustworthy and reliable and likely to stick around for a bit ;)

evan2645:
Turns out I had opened an issue on this same item some months ago and forgot about it :) #6 

The SPIFFE scheme is now registered with IANA: https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml

Thanks for the nudge @mhausenblas

PR #63, opened 2018-02-14
Update SPIFFE specification to directly reference RFC 3986

evan2645:
This commit is intended to 1) directly reference RFC 3986 so it is clear that the constraints defined there apply to a SPIFFE ID, and 2) be more specific about allowances defined in RFC 3986 that do not apply to SPIFFE IDs.

Additionally, update example SPIFFE IDs to use `example.com` instead of `acme.com`, as the former is preferred due to its reserved status. See section 6.5 of [RFC 6761](https://tools.ietf.org/html/rfc6761) for more information.

This work is related to #62.


justinburke (approved):
Pulled up the RFC to cross check... LGTM.

mattmoyer (approved):
Looks good to me.

PR #66, opened 2018-02-23
 Introduce the SPIFFE Workload Endpoint and API specifications

evan2645:
This commit introduces two separate specifications. The first (SPIFFE Workload Endpoint) codifies how to find and communicate with an identity agent, exposing a nominally node-local identity endpoint capable of delivering multiple services. The second defines the SPIFFE Workload API, a service which can issue workload identity, solving the problem known as identity bootstrapping, or trust bootstrapping. The latter is exposed through the former.

This work is a product of SPIFFE SIG-Specification - they have worked over the last two months developing this specification set. It has been written with extensibility in mind, and SIG-Specification expects additional services to be introduced in the near future. The SIG-Specification working documents can be found [here](https://docs.google.com/document/d/1XilXEa0C3uoZoYqqOfl3fR50TzuM2OfcpOAV6mJGHZ4).


standards/SPIFFE_Workload_API.md
```diff
@@ -0,0 +1,170 @@
...
+    // ASN.1 DER encoded certificate chain. MAY include intermediates,
+    // the leaf certificate (or SVID itself) MUST come first.
+    bytes x509_svid = 2;
+
+    // ASN.1 DER encoded PKCS#8 private key. MUST be unencrypted.
+    bytes x509_svid_key = 3;
+
+    // CA certificates belonging to the Trust Domain
+    // ASN.1 DER encoded
+    bytes bundle = 4;
```
evan2645, line 129: Placing the "native" bundle in the X509SVID message complicates validation, since the validator must understand when to consult this bundle or the federated bundles map. Feels like it makes more sense to put _all_ bundles in federated_bundles, including those belonging to issued SVIDs... thoughts? Maybe that means federated_bundles -> bundles?

justinburke: I had been assuming that this is basically the certificate chain that would be presented during a TLS handshake. Whereas the certificates in `federated_bundles` would be the self-signed roots of the foreign trust domain; at least for the purposes of TLS, including subordinate CA certs from a foreign trust domain isn't necessary since a foreign server will send those intermediate certs during the handshake.

If we're on the same page here about the intended usage of `bundle`, then perhaps we should add a comment to clarify (something like: `bundle` should include all intermediate certificates needed during a TLS handshake).

justinburke: To address your concern about the validator, we could add another field like `string trust_domain`. Not sure it's really justified though, a client should be able to pull the trust domain out of either `spiffe_id` or `x509_svid`.

standards/SPIFFE_Workload_API.md
```diff
@@ -0,0 +1,170 @@
...
+    // CA certificate bundles belonging to foreign Trust Domains that the
+    // workload should trust, keyed by the SPIFFE ID of the foreign
+    // domain. Bundles are ASN.1 DER encoded.
+    map<string, bytes> federated_bundles = 3;
+}
+
+// The X509SVID message carries a single SVID and all associated
+// information, including CA bundles.
+message X509SVID {
+    // The SPIFFE ID of the SVID in this entry
```
justinburke: The requirement that `spiffe_id` match the ID in `x509_svid` should (IMO) be strengthened. 
Could be clarified with something like: The SPIFFE ID of this SVID. Must match the SPIFFE ID contained in `spiffe_id`.

standards/SPIFFE_Workload_API.md
```diff
@@ -0,0 +1,170 @@
...
+// associated information. It also carries a set of global CRLs, and a
+// TTL to inform the workload when it should check back next.
+message X509SVIDResponse {
+    // A list of X509SVID messages, each of which includes a single
+    // SPIFFE Verifiable Identity Document, along with its private key
+    // and bundle.
+    repeated X509SVID svids = 1;
+
+    // ASN.1 DER encoded
+    repeated bytes crl = 2;
```
justinburke, line 105: Are the CRLs here issued by CAs in this SVID's same trust domain? Any trust domain? Or just CRLs covering this SVID (up the cert chain to this SVID's root)?

standards/SPIFFE_Workload_API.md
```diff
@@ -0,0 +1,170 @@
...
+
+    // CA certificates belonging to the Trust Domain
+    // ASN.1 DER encoded
+    bytes bundle = 4;
+
+}
+```
+
+
+All fields in the `X509SVID` message are mandatory, and MUST contain a non-default value. Clients receiving an `X509SVID` message in which any field has a default value SHOULD report an error and discard the message.
```
justinburke, line 135: Do we want to allow use of a Handshaker API or HSM/TPM to protect private keys? If so, can we make `x509_svid_key` optional?

standards/SPIFFE_Workload_API.md
evan2645: 👍 agreed

standards/SPIFFE_Workload_API.md
evan2645: >I had been assuming that this is basically the certificate chain that would be presented during a TLS handshake.

The chain (if any) is currently included in the `x509_svid` field (there is a comment there to that effect). This allows the service provider to control the presented chain. The `bundle` field must include the trust domain roots so that the service can authenticate presented certificates in its own trust domain, and if it additionally provided the intermediates, then it would be up to the workload to decide how many of the intermediates to present with the leaf... that feels like a slippery slope IMO.

>To address your concern about the validator, we could add another field like string trust_domain. 

My comment here on possibly moving the trust domain roots to `federated_bundles` is geared towards avoiding an extra if/else in client code. The way it is currently, client validation has to have something like:
```go
if spiffe.TrustDomain(clientSVID) == myTrustDomain {
     // Validate using certs from `bundle`
} else {
     // Validate using certs from `federated_bundle`
}
```
By having `federated_bundle` be the default (only) case, I think we can 1) slightly simplify the validation logic, and 2) lean towards federation being the norm rather than the exception.

standards/SPIFFE_Workload_API.md
evan2645: My thinking thus far has been that Handshaker API would be exposed over a dedicated gRPC service, separated from the X509SVID service. I'm under the impression that a Handshaker API would be able to accomplish TLS negotiation without necessarily tying itself to X509

If we protect a key pair with HSM/TPM, do you think we'd still use this X509SVID endpoint? Or perhaps we'd introduce a new service like Handshaker or something PKCS 11-shaped? I'm interested to hear more about this...

standards/SPIFFE_Workload_API.md
evan2645: I think the intention here is to deliver CRLs for any/all trust domains...

Another way could be to ship them as a map, keyed by the trust domain they belong to... though, there may be more than one CRL per trust domain.

I should note that we discussed the possibility that the CRL signatures need not be checked by the client when being received over the workload API because the workload API is already trusted to deliver identity and trust roots. As a side-effect of that, the door would be open for a workload's trust domain to ship CRLs for foreign issuers. Whether this is a good idea or not, I'm not sure :).

evan2645:
This PR is now stale and is blocking further SIG-Spec work. I will be merging Wednesday evening next week (July 11th) unless there are objections. If we wish to make modifications, we can do so with a targeted follow-on PR(s).

If you object to merging this, please leave a comment indicating why, along with a suggested correction. Thanks in advance

standards/SPIFFE_Workload_API.md
justinburke: Having a map from trust domain to CRL(s) would probably help out client implementors.

As for not checking the signatures of the CRLs: sounds like there would be an implicit contract that server is expected to validate signatures as needed for all of its workloads. Expecting clients not to validate signatures sounds suspect to me... but if that's the intention here, would recommend documenting it.

standards/SPIFFE_Workload_API.md
justinburke: Seems like we'd need to somehow indicate that the private key is intentionally mentioning and indicate that the Handshaker API is required. I think it's fine to tie to X509 here and (when support for non-exported key is needed) to add additional fields (ie. handshaker endpoint field or field for HSM key identity).

justinburke:
Thanks for pinging this review. I'll add a couple more comments for your consideration.

justinburke (dismissed):

standards/SPIFFE_Workload_API.md
evan2645: I opened https://github.com/spiffe/spiffe/issues/82 to track this

standards/SPIFFE_Workload_API.md
evan2645: Fixed up

standards/SPIFFE_Workload_API.md
evan2645: Given the current behavior, if an administrator has earmarked an identity for non-exported keys only, a workload hitting X509SVID endpoint would receive a gRPC permission denied error. The workload would have to "know" that it needs to speak to a different endpoint (e.g. handshaker api). I like the idea of having a field which indicates that the key is missing because it is not exportable, however I don't think that this will work if we are also using gRPC error code.

If it's OK with you, I think we should address this when drafting the handshaker spec. I think we can probably update this text in a way that retains compatibility with existing non-handshake-aware clients.

evan2645 (commented):
Thank you very much for the review @justinburke. Responded to a couple and opened a new issue. Gonna go ahead and merge this. Thanks again

PR #68, opened 2018-02-26
Clarify that X.509-SVID Signing Certificate may be an intermediate

evan2645:
There has been some confusion over whether or not SPIFFE IDs must be set from the root down. The intention of the current text is that only the most immediate signing certificate, which may be an intermediate, must have a SPIFFE ID. The issuer of the signing certificate is not required to be an SVID.

We should additionally revisit the MUST which mandates that signing certificates themselves be an SVID. I think that this makes sense logically, though the MUST was put in place at a time when the text described its use during the validation process - this is no longer the case. This matter probably warrants further discussion in SIG-Spec.

@banks does this suffice?

### DIFF
--- a/standards/X509-SVID.md
+++ b/standards/X509-SVID.md
@@ -46,7 +46,7 @@ Leaf certificate SPIFFE IDs MUST have a non-root path component. See [section 4.
 ### 3.2. Signing Certificates
 An X.509 SVID signing certificate is one which has set `keyCertSign` in the key usage extension. It additionally has the `CA` flag set to `true` in the basic constraints extension (see [section 4.1](#4.1.-basic-constraints)). That is to say, it is a CA certificate.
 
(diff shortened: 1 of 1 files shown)


banks:
Looks good to me - thanks!

I think if the wording can be changed to `should` then it will remove another obstacle for potential implementors who want to remain as widely compatible with external/existing PKI as possible (e.g. external CA that doesn't support URI SAN or has existing incompatible policies on how CA certificates are named). 

In practice, I suspect some (most?) implementations will choose not enforce this anyway in order to meet compatibility goals so allowing them to do so without violating spec seems like a good move unless there is a compelling technical reason not to.

evan2645:
Just going through and clearing out some stale PRs and issues. This PR no longer applies as signing certificates are not required to be an SVID anymore. Closing this out

PR #75, opened 2018-03-29
[#56] Clarify that Subject is not required for X.509 leaf SVIDs

### LINKED ISSUE #56: Clarify SAN usage in the X509-SVID specification

evan2645:
This commit updates X509-SVID to state that the Subject field is not required. If omitted, the SAN URI must be marked as critical per RFC 5280.

Fixes #56

### DIFF
--- a/standards/X509-SVID.md
+++ b/standards/X509-SVID.md
@@ -41,7 +41,7 @@ This section discusses the relationship between leaf, root, and intermediate cer
 ### 3.1. Leaf Certificates
 A leaf certificate is an SVID which serves to identify a caller or resource. They are signed by the signing authority of the trust domain in which they reside, and are suitable for use in authentication processes. A leaf certificate (as opposed to a signing certificate, [section 3.2](#3.2.-signing-certificates)) is the only type which may serve to identify a resource or caller.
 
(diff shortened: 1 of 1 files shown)


spikecurtis (approved):

spikecurtis:
@evan2645 I replaced an instance of "it," which was slightly ambiguous on my first read, to "Subject" which is slightly longer, but more clear.

evan2645:
👍 thanks for this @spikecurtis 

All SPIFFE repo's require DCO signoff - I can push this one through because the contribution is trivial, but something to remember in the future.

PR #81, opened 2018-05-29
Clarifies DCO contribution guidelines

ajessup:
Makes the documentation a little more explicit for folks who may not be familiar with the DCO process.

Signed-off-by: Andrew Jessup <ajessup@gmail.com>


evan2645 (approved):
Thanks for this @ajessup.

Perhaps we should also put a blurb here? https://github.com/spiffe/spire/blob/master/.github/PULL_REQUEST_TEMPLATE.md

PR #84, opened 2018-09-04
Update SPIFFE-ID.md

zparnold:
I believe this may have been a miss on the title of SPIFFE (but I'm brand new here so I could be wrong.)


evan2645 (approved):
Thank you @zparnold!

PR #86, opened 2018-10-02
JWT-SVID Specification

evan2645:
In typical use, SVIDs are backed by an asymmetric key pair, and verification of the identity is done by proving ownership of the private key. X509-SVIDs are frequently paired with mutually authenticated TLS in order to accomplish this.

Using mutually authenticated TLS as the proof of ownership mechanism works well for point-to-point communication, but can’t address use cases in which TLS is terminated anywhere other than the compute endpoint that a request is ultimately destined for. Layer 7 load balancers and proxies, in particular, suffer from this problem. As a result, the community is in need of a solution which can prove identity at Layer 7, allowing the assertion to survive traversal of Layer 7 boundaries. This specification defines the JWT SVID, which is designed to provide immediate value in solving difficulties associated with asserting identity across Layer 7 boundaries, complimenting the rest of the SPIFFE specification set.

### DIFF
--- a/standards/JWT-SVID.md
+++ b/standards/JWT-SVID.md
@@ -28,7 +28,7 @@ Appendix A. [Validation Reference](#appendix-a-validation-reference)
 ## 1. Introduction
 JWT-SVID is the first token-based SVID in the SPIFFE specification set. Aimed at providing immediate value in solving difficulties associated with asserting identity across Layer 7 boundaries, compatibility with existing applications and libraries is a core requirement.
 
-JWT-SVIDs are standard JWT tokens with a handful of restrictions applied. JWTs have historically proven difficult to get right, gaining a reputation in the security community as a technology which is likely to introduce vulnerabilities in its deployments and implementations. JWT-SVID takes steps to mitigate these problems as much as possible without breaking compatibility with existing applications and libraries.
+JWT-SVIDs are standard JOSE-protected JWT tokens with a handful of restrictions applied. JOSE has historically proven difficult to implement securely, gaining a reputation in the security community as a technology which is likely to introduce vulnerabilities in its deployments and implementations. JWT-SVID takes steps to mitigate these problems as much as possible without breaking compatibility with existing applications and libraries.
 
 ## 2. JOSE Header
 Historically, complexity introduced by the cryptographic agility of the JOSE header has led to a series of vulnerabilities in popular JWT implementations. To avoid such pitfalls, this specification greatly restricts the allowances originally afforded. This section describes the permitted registered headers, as well as their values. Any header not described here, registered or private, MUST NOT be included in the JWT-SVID JOSE Header.
@@ -59,7 +59,7 @@ This section outlines the requirements and restrictions placed upon existing reg
 The `sub` claim MUST be set to the SPIFFE ID of the workload to which it is issued. This is the primary claim against which workload identity is asserted.
 
 ### 3.2. Audience
-The `aud` claim MUST be present, containing one or more values. Validators MUST reject tokens without an `aud` claim set. It is strongly recommended that the number of values be limited to one in normal cases. Please see the Security Considerations section for more information.
+The `aud` claim MUST be present, containing one or more values. Validators MUST reject tokens without an `aud` claim set, or if the value that the validator identifies with is not present as an `aud` element. It is strongly recommended that the number of values be limited to one in normal cases. Please see the Security Considerations section for more information.
 
 The values chosen are site-specific, and SHOULD be scoped to the service which it is intended to be presented to. For example, `reports` or `spiffe://example.org/reports` are suitable values for tokens which are presented to the reports service. Values such as `production` or `spiffe://example.org/` are discouraged due to their wide scope, opening the possibility for impersonation if just a single service in `production` is compromised.
 
@@ -90,6 +90,8 @@ This section outlines the security considerations that implementers and users sh
 Being a bearer token, JWT-SVIDs are susceptible to replay attacks. By requiring that the `aud` and `exp` claims be set, this specification has taken steps to improve the situation, but is unable to solve it completely while retaining validation compatibility with [RFC 7515][1]. It is very important to understand this risk. Use of an aggressive value for the `exp` claim is recommended. Some users may wish to leverage the `jti` claim despite the added overhead. While use of the `jti` claim is permitted by this specification, it should be noted that JWT-SVID validators are not required to track `jti` uniqueness.
 
 ### 6.2. Audience
+There is an implicit trust granted to recipients of JWT-SVIDs. Tokens sent to one audience can be replayed to another audience should more than one be present. For example, if Alice has a token with audiences Bob and Chuck, and transmits that token to Chuck, then Chuck can impersonate Alice by sending the same token to Bob. As such, care should be taken when minting a JWT-SVID with more than one audience. Single audience JWT-SVID tokens are strongly recommended to limit the scope of replayability.
+
 There is an implicit trust granted to recipients of JWT-SVIDs. Tokens sent to one audience can be replayed to another audience should more than one be present. As such, care should be taken when minting a JWT-SVID with more than one audience. Single audience JWT-SVID tokens are strongly recommended to limit the scope of replayability.
 
 ### 6.3. Transport Security
--- a/standards/JWT-SVID.md
+++ b/standards/JWT-SVID.md
@@ -1,0 +1,112 @@
+# The JWT SPIFFE Verifiable Identity Document
+
+## Status of this Memo
+This document specifies an experimental identity document standard for the internet community, and requests discussion and suggestions for improvements. It is a work in progress. Distribution of this document is unlimited.
+
+## Table of Contents
+
+1\. [Introduction](#1-introduction)  
+2\. [JOSE HEader](#2-jose-header)  
+2.1. [Algorithm](#21-algorithm)  
+2.2. [Key ID](#22-key-id)  
+2.3. [Type](#23-type)  
+3\. [JWT Claims](#3-jwt-claims)  
+3.1. [Subject](#31-subject)  
+3.2. [Audience](#32-audience)  
+3.3. [Expiration Time](#33-expiration-time)  
+4\. [Token Signing and Validation](#4-token-signing-and-validation)  
+5\. [Token Transmission](#5-token-transmission)  
+5.1. [Serialization](#51-serialization)  
+5.2. [HTTP](#52-http)  
+5.3. [gRPC](#53-grpc)  
+6\. [Security Considerations](#6-security-considerations)  
+6.1. [Replay Protection](#61-replay-protection)  
+6.2. [Audience](#62-audience)  
+6.3. [Transport Security](#63-transport-security)  
+Appendix A. [Validation Reference](#appendix-a-validation-reference)  
+
+## 1. Introduction
+JWT-SVID is the first token-based SVID in the SPIFFE specification set. Aimed at providing immediate value in solving difficulties associated with asserting identity across Layer 7 boundaries, compatibility with existing applications and libraries is a core requirement.
+
+JWT-SVIDs are standard JWT tokens with a handful of restrictions applied. JWTs have historically proven difficult to get right, gaining a reputation in the security community as a technology which is likely to introduce vulnerabilities in its deployments and implementations. JWT-SVID takes steps to mitigate these problems as much as possible without breaking compatibility with existing applications and libraries.
+
+## 2. JOSE Header
+Historically, complexity introduced by the cryptographic agility of the JOSE header has led to a series of vulnerabilities in popular JWT implementations. To avoid such pitfalls, this specification greatly restricts the allowances originally afforded. This section describes the permitted registered headers, as well as their values. Any header not described here, registered or private, MUST NOT be included in the JWT-SVID JOSE Header.
+
+Only JWS is supported, and all header values MUST reside in a JWS Protected Header.
+
+### 2.1. Algorithm
+The `alg` header MUST be set to `RS256`.
+
+Most vulnerabilities in JWT come from mixing and matching of cryptographic techniques, so JWT-SVID defines only a single technique as valid. By taking this approach, we aim to sidestep the pitfalls associated with cryptographic agility. In the event that the above-defined technique needs to be changed, it will be done by releasing a new version of JWT-SVID.
+
+The `RS256` algorithm was chosen for a number of reasons. First, it is a goal of this specification to remain compatible with existing JWT libraries and applications. This largely restricts options to the algorithms defined in [RFC 7518 Section 3.1][5], ruling out the use of more efficient algorithms like those based on the Ed25519 curve.
+
+Second, the `RS256` algorithm is very widely implemented and supported by practically all JWT libraries and consumers that are capable of asymmetric token signing and validation. Finally, the RSA algorithm is less likely to suffer from critical implementation flaws than those based on elliptic curves.
+
+### 2.2. Key ID
+The `kid` header is optional.
+
+### 2.3. Type
+The `typ` header is optional. If set, its value MUST be either `JWT` or `JOSE`.
+
+## 3. JWT Claims
+The JWT-SVID specification does not introduce any new claims, though it does set some restrictions on the registered claims defined by [RFC 7519][1]. Registered claims not described in this document, in addition to private claims, MAY be used as implementers see fit. It should be noted, however, that reliance on claims which are not defined here may impact interoperability, as the producing and consuming applications must independently agree. Implementers should exercise caution when introducing additional claims and carefully consider the impact on SVID interoperability, particularly in environments where the implementer does not control both the producer and the consumer.
+
+This section outlines the requirements and restrictions placed upon existing registered claims by the JWT-SVID specification.
+
+### 3.1. Subject
+The `sub` claim MUST be set to the SPIFFE ID of the workload to which it is issued. This is the primary claim against which workload identity is asserted.
+
+### 3.2. Audience
+The `aud` claim MUST be present, containing one or more values. Validators MUST reject tokens without an `aud` claim set. It is strongly recommended that the number of values be limited to one in normal cases. Please see the Security Considerations section for more information.
+
+The values chosen are site-specific, and SHOULD be scoped to the service which it is intended to be presented to. For example, `reports` or `spiffe://example.org/reports` are suitable values for tokens which are presented to the reports service. Values such as `production` or `spiffe://example.org/` are discouraged due to their wide scope, opening the possibility for impersonation if just a single service in `production` is compromised.
+
+### 3.3. Expiration Time
+The `exp` claim MUST be set, and validators MUST reject tokens without this claim. Implementers are encouraged to keep the validity period as small as is reasonably possible, however this specification does not set any hard upper limits on its value.
+
+## 4. Token Signing and Validation
+JWT-SVID signing and validation semantics are the same as regular JWTs. Validators MUST ensure that the `alg` header is set to `RS256` before processing.
+
+JWT-SVID signatures are computed and validated following the steps outlined in [RFC 7515 section 7][2]. The `aud` and `exp` claims MUST be present and processed according to [RFC 7519][1] sections [4.1.3][3] and [4.1.4][4]. Validators receiving tokens without the `aud` and `exp` claims set MUST reject the token.
+
+## 5. Token Transmission
+This section describes the manner in which a JWT-SVID may be transmitted from one workload to another.
+
+### 5.1. Serialization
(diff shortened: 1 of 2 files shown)


standards/JWT-SVID.md
```diff
@@ -0,0 +1,112 @@
...
+
+## 2. JOSE Header
+Historically, complexity introduced by the cryptographic agility of the JOSE header has led to a series of vulnerabilities in popular JWT implementations. To avoid such pitfalls, this specification greatly restricts the allowances originally afforded. This section describes the permitted registered headers, as well as their values. Any header not described here, registered or private, MUST NOT be included in the JWT-SVID JOSE Header.
+
+Only JWS is supported, and all header values MUST reside in a JWS Protected Header.
+
+### 2.1. Algorithm
+The `alg` header MUST be set to `RS256`.
+
+Most vulnerabilities in JWT come from mixing and matching of cryptographic techniques, so JWT-SVID defines only a single technique as valid. By taking this approach, we aim to sidestep the pitfalls associated with cryptographic agility. In the event that the above-defined technique needs to be changed, it will be done by releasing a new version of JWT-SVID.
```
evan2645: Request for public comment on this point. Will imposing this restriction do more harm than good by hindering adoption in projects which support configurable key types and sizes?

sarahscheffler: The most important thing is for the sent header to match what the recipient expects.  The [classic thing to do](https://www.chosenplaintext.ca/2015/03/31/jwt-algorithm-confusion.html) is to replace the `alg` value with `none` and bypass the check entirely.  As @gcallaghan mentioned elsewhere, simply requiring that the field be an asymmetric algorithm (RS256, RS384, RS512, ES256, ES384, ES512) is likely sufficient.  The best thing would be for each receiver to accept exactly one of these, and verify that it was there before accepting.  But there's no reason that different projects couldn't use different sorts of keys.  The various JWT libraries should have functionalities to check this, but [many still don't](https://jwt.io/).

standards/JWT-SVID.md
```diff
@@ -0,0 +1,112 @@
...
+
+The `RS256` algorithm was chosen for a number of reasons. First, it is a goal of this specification to remain compatible with existing JWT libraries and applications. This largely restricts options to the algorithms defined in [RFC 7518 Section 3.1][5], ruling out the use of more efficient algorithms like those based on the Ed25519 curve.
```
sarahscheffler: Lots of JWT libraries support elliptic curves, and some don't even support RSA.  I agree if you're trying to maximize your compatibility, requiring RSA will do that, so it's a tradeoff between being compatible with more JWT libraries and supporting better/more algorithms.

standards/JWT-SVID.md
```diff
@@ -0,0 +1,112 @@
...
+
+Second, the `RS256` algorithm is very widely implemented and supported by practically all JWT libraries and consumers that are capable of asymmetric token signing and validation. Finally, the RSA algorithm is less likely to suffer from critical implementation flaws than those based on elliptic curves.
+
+### 2.2. Key ID
+The `kid` header is optional.
```
sarahscheffler, line 56: The `kid` header should be mandatory for cases where the recipient does not have another way of identifying the correct key to use (when there are multiple key options).

standards/JWT-SVID.md
```diff
@@ -0,0 +1,112 @@
...
+
+### 2.3. Type
+The `typ` header is optional. If set, its value MUST be either `JWT` or `JOSE`.
+
+## 3. JWT Claims
+The JWT-SVID specification does not introduce any new claims, though it does set some restrictions on the registered claims defined by [RFC 7519][1]. Registered claims not described in this document, in addition to private claims, MAY be used as implementers see fit. It should be noted, however, that reliance on claims which are not defined here may impact interoperability, as the producing and consuming applications must independently agree. Implementers should exercise caution when introducing additional claims and carefully consider the impact on SVID interoperability, particularly in environments where the implementer does not control both the producer and the consumer.
+
```
sarahscheffler, line 63: The most important thing from a security standpoint is to be consistent.  Pick a set of headers and claims that you're always going to use, and then always validate each claim to make sure it matches what you expected.

standards/JWT-SVID.md
```diff
@@ -0,0 +1,112 @@
...
+This section outlines the requirements and restrictions placed upon existing registered claims by the JWT-SVID specification.
+
+### 3.1. Subject
+The `sub` claim MUST be set to the SPIFFE ID of the workload to which it is issued. This is the primary claim against which workload identity is asserted.
+
+### 3.2. Audience
+The `aud` claim MUST be present, containing one or more values. Validators MUST reject tokens without an `aud` claim set. It is strongly recommended that the number of values be limited to one in normal cases. Please see the Security Considerations section for more information.
```
sarahscheffler: Validators should reject tokens that are sent to the wrong audience.  (As currently worded, it only requires checking that the `aud` claim is present, without verifying what it is.)

sarahscheffler: The same goes for the `iss` field for any clients that share keys, so you know who sent the token.  (Hopefully there are no/few cases where clients share keys.)

standards/JWT-SVID.md
```diff
@@ -0,0 +1,112 @@
...
+
+### 5.3. gRPC
+The gRPC protocol uses HTTP/2. As a result, the HTTP transmission guidelines in the [HTTP section](#52-http) equally apply. Concretely, gRPC implementations SHOULD set a metadata key `authorization` with a value of `Bearer <serialized_token>`.
+
+## 6. Security Considerations
+This section outlines the security considerations that implementers and users should take into account when using JWT-SVID.
+
+### 6.1. Replay Protection
+Being a bearer token, JWT-SVIDs are susceptible to replay attacks. By requiring that the `aud` and `exp` claims be set, this specification has taken steps to improve the situation, but is unable to solve it completely while retaining validation compatibility with [RFC 7515][1]. It is very important to understand this risk. Use of an aggressive value for the `exp` claim is recommended. Some users may wish to leverage the `jti` claim despite the added overhead. While use of the `jti` claim is permitted by this specification, it should be noted that JWT-SVID validators are not required to track `jti` uniqueness.
```
sarahscheffler, line 98: If you want to add the main idea of how to do this, even though it's infeasible for most cases where you would want to use tokens in the first place: To fully protect against replay attacks by using `jti`, the validator can keep state and reject duplicate `jti` values, until the original token's `exp` time has passed.

standards/JWT-SVID.md
```diff
@@ -0,0 +1,112 @@
...
+
+### 6.2. Audience
+There is an implicit trust granted to recipients of JWT-SVIDs. Tokens sent to one audience can be replayed to another audience should more than one be present. As such, care should be taken when minting a JWT-SVID with more than one audience. Single audience JWT-SVID tokens are strongly recommended to limit the scope of replayability.
```
sarahscheffler: To clarify: If more than one _validator_ is considered to be the same `aud`, then the tokens **can** be replayed between these two validators because the exact same token will be accepted by both.

If there are two validators with _different_ `aud` values, and the `aud` value is validated, then the tokens **cannot** be replayed between these two validators, because either the `aud` value will reject for the second validator, or the signature will fail for the second validator.  (Or the attacker has the signing key in which case there are bigger problems.)

sarahscheffler (commented):
I have two major comments, as well as a handful of minor ones.

1. On the security of the `alg` header - Security-wise, each application should pick one (asymmetric) algorithm and stick with it.  But the SVID spec need not restrict everyone to using the same algorithm and key size.  I second Grant's (will @ if I get his github username) suggestion of requiring (and validating) that the `alg` field be set to one of the acceptable algorithms, but not mandating `RS256`.  See lines 41 and 43.

2. Clarified a claim about replay for multiple `aud` fields.  Having multiple validators use the same `aud` allows reply, having multiple `aud` values does not.  See line 93.

I also made a few other minor comments and clarifications inline.

standards/JWT-SVID.md
```diff
@@ -0,0 +1,112 @@
...
+6\. [Security Considerations](#6-security-considerations)  
+6.1. [Replay Protection](#61-replay-protection)  
+6.2. [Audience](#62-audience)  
+6.3. [Transport Security](#63-transport-security)  
+Appendix A. [Validation Reference](#appendix-a-validation-reference)  
+
+## 1. Introduction
+JWT-SVID is the first token-based SVID in the SPIFFE specification set. Aimed at providing immediate value in solving difficulties associated with asserting identity across Layer 7 boundaries, compatibility with existing applications and libraries is a core requirement.
+
+JWT-SVIDs are standard JWT tokens with a handful of restrictions applied. JWTs have historically proven difficult to get right, gaining a reputation in the security community as a technology which is likely to introduce vulnerabilities in its deployments and implementations. JWT-SVID takes steps to mitigate these problems as much as possible without breaking compatibility with existing applications and libraries.
```
evan2645: Ported comment by @gcallaghan from draft spec:
```This may be slightly contentious. I would argue that the JOSE spec as written has been difficult to implement securely due to some ambiguity and the "crypto agility" as mentioned below. JWT's themselves are just a signed bundle of metadata with some reserved claims.```

File: standards/JWT-SVID.md [hunk 9]
```diff
@@ -0,0 +1,112 @@
...
+
+## 2. JOSE Header
+Historically, complexity introduced by the cryptographic agility of the JOSE header has led to a series of vulnerabilities in popular JWT implementations. To avoid such pitfalls, this specification greatly restricts the allowances originally afforded. This section describes the permitted registered headers, as well as their values. Any header not described here, registered or private, MUST NOT be included in the JWT-SVID JOSE Header.
+
+Only JWS is supported, and all header values MUST reside in a JWS Protected Header.
+
+### 2.1. Algorithm
+The `alg` header MUST be set to `RS256`.
```
evan2645: Ported comment by @gcallaghan from draft spec:
```I understand the reason to RECOMMEND RS256 and to REQUIRE an asymmetric alg, but this seems a little too restrictive. I know before NIST Suite B got thrown out there were some institutions that required ES384 or higher. One of the aspects of SPIFFE that is appealing is some flexibility in implementation.```

File: standards/JWT-SVID.md (hunk 1 above)
Comment (evan2645) [reply standards/JWT-SVID.md]: Ported comment by @gcallaghan from draft spec:
```OIDC recommends collision resistant names for custom claims https://openid.net/specs/openid-connect-core-1_0.html#AdditionalClaims, in practice implementers accomplish this by using a URI namespace https://auth0.com/docs/api-auth/tutorials/adoption/scope-custom-claims#custom-claims. This might be a worthwhile addition to the spec. Also, it might be nice to have guidance for some .well_known url for a consumer to advertise supported claims.```

File: standards/JWT-SVID.md (hunk 4 above)
Comment (evan2645) [reply standards/JWT-SVID.md:63]: Ported comment by @gcallaghan from draft spec:
```OIDC recommends collision resistant names for custom claims https://openid.net/specs/openid-connect-core-1_0.html#AdditionalClaims, in practice implementers accomplish this by using a URI namespace https://auth0.com/docs/api-auth/tutorials/adoption/scope-custom-claims#custom-claims. This might be a worthwhile addition to the spec. Also, it might be nice to have guidance for some .well_known url for a consumer to advertise supported claims.```

File: standards/JWT-SVID.md (hunk 1 above)
Comment (evan2645) [reply standards/JWT-SVID.md]: I think the hard part here is ensuring that validators actually take this extra step. Being rigid on alg was intended to be a measure which 1) makes it difficult to ignore, and 2) makes it easy to implement a check.

There have been several comments on this point... there are also many projects out there which allow varying key types, and how those projects might adopt this spec is a big question mark as long as this restriction is in place. I think the key here is, how can we be loose enough that folks have freedom to make choices but not so loose that they can make bad choices (or oversights!).

Perhaps this can be accomplished by simply forbidding the NONE alg... though, I have also heard of attacks which involve swapping an asymmetric alg for a symmetric one. IMO, if we tell validators that the alg value must be one of six different options, they are unlikely to implement the check.

If folks think that this requirement generally imposes more harm than good (that is what I seem to be hearing at the moment?) then we should definitely loosen it... though it would be prudent to consider how we might loosen it while also making it easy on validators so they don't cut corners.

Comment (evan2645) [reply standards/JWT-SVID.md]: @gcallaghan

File: standards/JWT-SVID.md (hunk 3 above)
Comment (evan2645) [reply standards/JWT-SVID.md:56]: @sarahscheffler is there danger in trying multiple keys?

This is currently optional because there are deployments in which there is always a single key only, and such deployments would probably end up with some arbitrary and meaningless key id... I suppose that doesn't hurt? Though requiring it in such instances felt a bit heavy handed.

File: standards/JWT-SVID.md (hunk 4 above)
Comment (evan2645) [reply standards/JWT-SVID.md:63]: I think we should steer clear of defining claims beyond the "standard" ones described here. There's a myriad of use cases and capturing them all in a generic way will be hard. Additionally, I'm personally a little hesitant to include arbitrary claims/metadata on identity documents due to interoperability concerns (which is the reason that this snippet of text was added).

I added a validation reference appendix at the end to try and make it clear which claims must be present, and how to validate them... reviewing this section, I can see that it's kind of hard to make that jump. Additionally, all the claims described here are mandatory. Perhaps we can add some text to this section highlighting the fact that all claims described here are mandatory, and/or insert a direct reference to the appendix?

File: standards/JWT-SVID.md (hunk 5 above)
Comment (evan2645) [reply standards/JWT-SVID.md]: Good catch, I'll fix this.

File: standards/JWT-SVID.md (hunk 8 above)
Comment (evan2645) [reply standards/JWT-SVID.md]: Good point, I will update the text to clarify this. Thanks for the feedback

File: standards/JWT-SVID.md (hunk 7 above)
Comment (evan2645) [reply standards/JWT-SVID.md]: > If there are two validators with different aud values, and the aud value is validated, then the tokens cannot be replayed between these two validators, because either the aud value will reject for the second validator, or the signature will fail for the second validator. (Or the attacker has the signing key in which case there are bigger problems.)

I'm not sure I follow this point... the key in use represents the subject, and not a relationship between the subject and the audience. Thus, the signature is valid for both audience values. If subject A cuts a token for audience X and Y, and sends that token to X, then X can replay the token to Y. Perhaps I am missing something?

File: standards/JWT-SVID.md (hunk 1 above)
Comment (gcallaghan) [reply standards/JWT-SVID.md]: I think the easiest way to help validators conform is to provide a schema for them to match. Since JWT is a JSON document it makes sense to  use a JSON Schema. An example of one that would take this spec and the comments into consideration would be:

```yaml
$id: "https://spiffe.com/jwt-svd.json"
$schema: "http://json-schema.org/draft-07/schema#"
description: "SPIFFE schema to validate JWT SVD"
type: "object"
properties:
  header:
    type: "object"
    properties:
      alg:
        type: "string"
        description: |
          The alg (algorithm) Header Parameter identifies the cryptographic
          algorithm used to secure the JWS.  The JWS Signature value is not
          valid if the alg value does not represent a supported algorithm or
          if there is not a key for use with that algorithm associated with the
          party that digitally signed or MACed the content. The alg value is a case-
          sensitive ASCII string containing a StringOrURI value. This Header
          Parameter MUST be present and MUST be understood and processed by
          implementations.
          see https://tools.ietf.org/html/rfc7515#section-4.1.1
        enum:
          - "RS256"
          - "RS384"
          - "RS512"
          - "ES256"
          - "ES384"
          - "ES512"
      kid:
        type: "string"
        description: |
          The kid (key ID) Header Parameter is a hint indicating which key
          was used to secure the JWS.  This parameter allows originators to
          explicitly signal a change of key to recipients.  The structure of
          the kid value is unspecified.  Its value MUST be a case-sensitive
          string.  Use of this Header Parameter is OPTIONAL.
          see https://tools.ietf.org/html/rfc7515#section-4.1.4
    required:
      - "alg"
  payload:
    type: "object"
    properties:
      aud:
        type: "string"
        description: |
          A JSON string value, with the additional requirement that while
          arbitrary string values MAY be used, any value containing a ":"
          character MUST be a URI [RFC3986].  StringOrURI values are
          compared as case-sensitive strings with no transformations or
          canonicalizations applied.
          see https://tools.ietf.org/html/rfc7519#section-4.1.3
      exp:
        type: "integer"
        description: |
          The "exp" (expiration time) claim identifies the expiration time on
          or after which the JWT MUST NOT be accepted for processing.  The
          processing of the "exp" claim requires that the current date/time
          MUST be before the expiration date/time listed in the "exp" claim.
          Implementers MAY provide for some small leeway, usually no more than
          a few minutes, to account for clock skew.  Its value MUST be a number
          containing a NumericDate value.  Use of this claim is OPTIONAL.
          see https://tools.ietf.org/html/rfc7519#section-4.1.4
    required:
      - "aud"
      - "exp"
    additionalProperties: false
    patternProperties:
      \w+:(\/?\/?)[^\s]+:
        type: "string"
```

Passing SVD

```json
{
  "header": {
    "alg": "RS256"
  },
  "payload": {
    "exp": 1234,
    "aud": "http:example.com",
    "https://wat.io": "bar",
    "arn:aws:a4b:region:accountid:resourcetype/resource": "foo"
  }
}
```

Failing SVD

```json
{
  "header": {
    "alg": "RS256"
  },
  "payload": {
    "exp": 1234,
    "aud": "http:example.com",
    "wat": "bar"
  }
}
```

Error Message

```bash
$ ajv -s schema.json -d svd.fail.json
svd.fail.json invalid
[ { keyword: 'additionalProperties',
    dataPath: '.payload',
    schemaPath: '#/properties/payload/additionalProperties',
    params: { additionalProperty: 'wat' },
    message: 'should NOT have additional properties' } ]
```

standards/JWT-SVID.md
gcallaghan: Yeah, what I was trying to convey is I wouldn't want to exclude implementors from including their own custom claims. I would suggest that if custom claims are added they SHOULD/MUST be collision resistant by using a URI namespace.

standards/JWT-SVID.md
gcallaghan: In practice, `jti` is usually used as a blacklist to manually invalidate tokens rather than requiring a single token per request which could be onerous. This blacklist contains the `exp` as well so it can be purged routinely.

standards/JWT-SVID.md
mcpherrinm: Requiring RSA to the exclusion of elliptic curve algorithms seems like a pretty big negative to me.  While alg none is obviously dangerous, Elliptic curves offer a number of advantages over RSA, most notably in size, which can be nice in contexts when you're counting packets over the wire.

fwiw, github.com/square/go-jose simply does not implement alg: none.

standards/JWT-SVID.md
sarahscheffler: Generically, the thing I was worried about are "[related key attacks](https://eprint.iacr.org/2015/1135.pdf)" which occur when two keys accessible to the attacker are related in some exploitable way.  It is surprisingly easy to mess up the independence of keys!  (See [here](https://eprint.iacr.org/2012/064.pdf), though that particular flaw would imply much bigger problems than just what I mentioned before.)  So, from a theoretical perspective, I recommend trying not to misalign the keys.  But from a practical one, I don't know of any concrete attacks of this type against RSA or ECDSA, and anyway the best defense against this is to have independent,. unrelated keys in the first place.

standards/JWT-SVID.md
sarahscheffler: Ah, I misunderstood, I had two concerns, one of which is now addressed, and the other of which is probably a non-issue:
1. I thought you were worried about a token from sender A with (only) aud X being mistakenly accepted by aud Y.  (As you mentioned, this doesn't make that much sense for a token, and also Y should reject anyway because it's not in the aud of the token.)
2. I also thought you were worried about a token from sender A with aud X being replayed _among members of audience X_.  I don't know how much of a concern this is, but if it is one, it's not addressed.

Assuming my second concern is a non-issue, I favor changing the text to clarify this for those who make the same misunderstanding:

> There is an implicit trust granted to recipients of JWT-SVIDs. A token from sender A to audiences X and Y can be replayed by audience X to audience Y.

standards/JWT-SVID.md
evan2645: Ah got it... I will add clarifying text here

standards/JWT-SVID.md
evan2645: 👍 I'll add the clarification - thanks for this

spikecurtis:
@evan2645 I've just noticed that there is language in the SPIFFE ID spec that will need updating to cover JWT SVIDs.  Relevant section is https://github.com/spiffe/spiffe/blob/master/standards/SPIFFE-ID.md#3-spiffe-verifiable-identity-document

Basically, that doc says that an SVID needs to contain a public key and in order to use an SVID the workload must be able to prove it owns the corresponding private key.  This doesn't apply to bearer token SVIDs like JWT, so the language needs to be updated to be more broad.

spikecurtis (approved):

PR #91, opened 2018-10-31
Clarify section 3.1 of X509-SVID

evan2645:
This commit is intended to clarify the fact that the signing authority
need not reside in the trust domain in which it is issuing leaf
certificates.

@justinburke does this suffice? If not, perhaps you could suggest a snippet to make it clearer?


justinburke:
Made a few edits. I think these changes enable us to allow a leaf to be signed by any signing cert. These changes make SPIFFE more like general PKI in that we don't have a strict naming relationship between leaf and signer. I think that's okay, but perhaps does make it even more important that we look at enhanced validation spec (like the tuples that we've discussed).

evan2645:
👍 LGTM

@mattmoyer since both Justin and I have commits on this PR, do you mind giving this a quick look?

spikecurtis (approved):

evan2645:
Thanks @spikecurtis!

PR #92, opened 2018-11-15
Added note about SPIFFE URIs that trailing slashes are disallowed

mweissbacher:
(Based on today's call)

Signed-off-by: Michael Weissbacher <mweissbacher@squareup.com>

### DIFF
--- a/standards/SPIFFE-ID.md
+++ b/standards/SPIFFE-ID.md
@@ -47,7 +47,7 @@ The trust domain is defined as the authority component of the URI - specifically
 ### 2.2. Path
 The path component of a SPIFFE ID allows for the unique identification of a given workload. The meaning behind the path is left open ended and the responsibility of the administrator to define.
 
-Paths MAY be hierarchical - similar to filesystem paths. The specific meaning of paths is reserved as an exercise to the implementer and are outside the SVID specification. However, to reduce ambiguity paths must not end with a trailing `/`. Some examples and conventions are expressed below.
+Paths MAY be hierarchical - similar to filesystem paths. The specific meaning of paths is reserved as an exercise to the implementer and are outside the SVID specification. However, to reduce ambiguity, paths should not end with a trailing `/`. Some examples and conventions are expressed below.
 
 * Identifying services directly
 
--- a/standards/SPIFFE-ID.md
+++ b/standards/SPIFFE-ID.md
@@ -47,7 +47,7 @@ The trust domain is defined as the authority component of the URI - specifically
 ### 2.2. Path
 The path component of a SPIFFE ID allows for the unique identification of a given workload. The meaning behind the path is left open ended and the responsibility of the administrator to define.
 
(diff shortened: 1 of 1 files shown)


mattmoyer (dismissed):

mcpherrinm:
in RFC-2119 speak, should this be a "SHOULD NOT", instead of a "MUST NOT"?

4. SHOULD NOT   This phrase, or the phrase "NOT RECOMMENDED" mean that
   there may exist valid reasons in particular circumstances when the
   particular behavior is acceptable or even useful, but the full
   implications should be understood and the case carefully weighed
   before implementing any behavior described with this label.

justinburke:
Hi. Is there a specific problem we're trying to address with this PR? I understand that trailing slashes could be ambiguous, but we've largely been leaving the semantics of the path component up to the implementation (and that's explicitly codified in the spec). So might make sense to have this as a recommendation, but not a hard requirement. I think my preference would be to use 'SHOULD NOT' if this is to be included at all.

mcpherrinm:
I think our main concern here is that what the definition of "hierarchical" means isn't clear enough to implement access control uniformly.  The spec currently mentions slash separated items, but doesn't make mention of trailing slashes.  We think this is potentially a compatibility hazard for implementers of access control libraries that use SPIFFE hierarchical paths.  By not recommending trailing slashes, there's a better chance that various workloads will be compatible.

mweissbacher:
> Hi. Is there a specific problem we're trying to address with this PR? I understand that trailing slashes could be ambiguous, but we've largely been leaving the semantics of the path component up to the implementation (and that's explicitly codified in the spec). So might make sense to have this as a recommendation, but not a hard requirement. I think my preference would be to use 'SHOULD NOT' if this is to be included at all.

If all certificates are issued by the same CA they would likely all look the same way, one way or the other. For the federation use case this could lead to mixed issuing and mixed ACLs. There is no standardized way of performing ACL checks currently. As this is an early standard this seemed like a good opportunity to reduce ambiguity.

justinburke:
Makes sense to anticipate these sorts of compatibility problems. Given that path components "MAY be hierarchical" and that the semantics of the path component is open ended, I propose we change this PR from "must not" to "SHOULD NOT".

standards/SPIFFE-ID.md
```diff
@@ -47,7 +47,7 @@ The trust domain is defined as the authority component of the URI - specifically
 ### 2.2. Path
 The path component of a SPIFFE ID allows for the unique identification of a given workload. The meaning behind the path is left open ended and the responsibility of the administrator to define.
 
-Paths MAY be hierarchical - similar to filesystem paths. The specific meaning of paths is reserved as an exercise to the implementer and are outside the SVID specification. However some examples and conventions are expressed below.
+Paths MAY be hierarchical - similar to filesystem paths. The specific meaning of paths is reserved as an exercise to the implementer and are outside the SVID specification. However, to reduce ambiguity paths must not end with a trailing `/`. Some examples and conventions are expressed below.
```
justinburke: Missing comma after "ambiguity":   However, to reduce ambiguity, paths ...

mattmoyer (dismissed):

spikecurtis (approved):

PR #118, opened 2020-01-23
Clarify JWT-SVID serialization requirements

### LINKED ISSUE #111: JWT - Compact Serialization
https://github.com/spiffe/spiffe/blob/master/standards/JWT-SVID.md
> **Section 5.1**
> 
> JWT-SVIDs MUST be serialized using the JWS Compact Serialization method described in RFC 7515 Section 3.1. Note that this precludes the use of a JWS Unprotected Header, as mandated in the JOSE Header section.

The above section seems to be redundant. A JWT is always compact serialized. When we say a JWT, that it self implies it is compact serialized (with JWS Compact Serialization or JWE Compact Serialization).

evan2645:
@prabath has pointed out that our requirement for JWT-SVIDs to be
serialized using JWS Compact Serialization is redundant because the JWT
RFC itself does not permit JWS JSON Serialization.

This commit attempts to clarify this point by amending section 5.1 with
text that states that the JWT RFC mandates this already. @prabath has
suggested that section 5.1 be struck in its entirety, however due
diligence by SIG-Spec has shown that there are a handful of popular
libraries that still permit this serialiazation format, thus we thought
it prudent for the comment to remain.

Some minor redundancy has been left in place to aid readers who do not
read the document in its entirety.

Fixes #111 


azdagron (approved):
This looks good to me!

PR #122, opened 2020-04-30
Clarify missing `use` paramerter

### LINKED ISSUE #121: SPIFFE Bundle Format: consider to make more clear the requirements for the `use` parameter
[Section 4.2.2](https://github.com/spiffe/spiffe/blob/master/standards/SPIFFE_Trust_Domain_and_Bundle.md#422-public-key-use) of the SPIFFE Trust Domain and Bundle specification has the requirements for the `use` parameter in the SPIFFE Bundle Format.
While implementing a SPIFFE Bundle parser, it was not completely clear how this parameter should be treated in conformance with the standard.
In particular, the specification says:
>The `use` parameter MUST be set.

and also:

> Clients encountering unknown `use` values MUST ignore the entire JWK element.

Which does not provide clarity for the case where there is a document that has some keys with valid SPIFFE values for the `use` parameter and some with no `use` specified.
It would be good if it is clear from the standard perspective how such case should be treated by software that produces/consumes documents using the SPIFFE Bundle Format.

spikecurtis:
Fixes #121


standards/SPIFFE_Trust_Domain_and_Bundle.md
```diff
@@ -84,7 +84,7 @@ Implementers SHOULD NOT include parameters which are defined neither here nor in
 The `kty` parameter MUST be set, and its behavior follows [Section 4.1][6] of RFC 7517. Clients encountering an unknown key type MUST ignore the entire JWK element.
 
 #### 4.2.2. Public Key Use
-The `use` parameter MUST be set. Its value indicates the type of identity document (or SVID) that it is authoritative for. At the time of this writing, only two SVID types are supported: `x509-svid` and `jwt-svid`. The values are case sensitive. Please see the respective SVID specifications for more information about `use` values. Clients encountering unknown `use` values MUST ignore the entire JWK element.
+The `use` parameter MUST be set. Its value indicates the type of identity document (or SVID) that it is authoritative for. At the time of this writing, only two SVID types are supported: `x509-svid` and `jwt-svid`. The values are case sensitive. Please see the respective SVID specifications for more information about `use` values. Clients encountering either a missing `use` parameter or unknown `use` value MUST ignore the entire JWK element.
```
justinburke: s/or unknown/or an unknown/

justinburke (changes requested):
One minor request below, otherwise LGTM.

azdagron (commented):
Also LGTM, pending Justin's suggestion.

spikecurtis:
@azdagron @justinburke fixed.

justinburke (approved):

PR #123, opened 2020-04-30
Clarify empty bundles

spikecurtis:
Clarifies that empty bundles are legal, discusses what to do, and how they are subtly different from bundles which contain no valid JWKs

Relates to #121


standards/SPIFFE_Trust_Domain_and_Bundle.md
```diff
@@ -75,6 +75,8 @@ The parameter `spiffe_refresh_hint` SHOULD be set. The refresh hint indicates ho
 #### 4.1.3. Keys
 The parameter `keys` MUST be present. Its value is an array of JWKs. Clients encountering unknown key types or uses MUST ignore the corresponding JWK element. Please see [Section 5][5] of RFC 7517 for more information about the semantics of the `keys` parameter.
 
+The `keys` parameter may contain an empty array. Clients may also encounter bundles which after processing yield no usable keys (i.e. no JWKs pass validation described below), and are effectively empty.  In both cases, workloads MUST treat all SVIDs from the trust domain as invalid.  SPIFFE control planes that process bundles MAY wish alert operators when they encounter actually empty or effectively empty bundles. The former indicates that the trust domain is no longer issuing SPIFFE identities, while the latter may indicate that the trust domain has migrated to a new key type or use not understood by the client.
+
```
justinburke: Some wordsmithing below. I tried to separate out the two types of empty array cases that you mentioned to more clearly describe the meaning and handling. While I agree that control plane implementations may want to raise an alert when encountering an empty key array (particularly if the key array were previously non-empty), I am not sure that it fits into this spec since the focus here is on the bundle itself and may be too prescriptive.

```suggestion
The `keys` parameter may contain an empty array. A trust domain which publishes an empty key array indicates that the trust domain has revoked any previously-published keys; workloads MUST treat all SVIDs from the trust domain as invalid and untrusted. Clients may also encounter bundles which, after processing, yield no usable keys (i.e. no JWKs pass the validation procedure described below) and are therefore effectively empty; a key array empty due to client-side processing may indicate that the trust domain is using key types not supported by the client.
```

justinburke (changes requested):

standards/SPIFFE_Trust_Domain_and_Bundle.md
spikecurtis: I wanted to indicate that in both cases of the empty array, workloads MUST stop accepting SVIDs from that trust domain.

We can drop the text about alerts; I agree that it's not the focus of this spec.

spikecurtis: @justinburke PTAL at the new text

justinburke (approved):

PR #142, opened 2020-11-03
Add TSC call details to the SPIFFE README

evan2645:
Now that the TSC calls are public, they should be documented on the SPIFFE readme. This PR does that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ SPIFFE is hosted by the [Cloud Native Computing Foundation](https://cncf.io) (CN
 * [CONTRIBUTING](/CONTRIBUTING.md)
 * [GOVERNANCE](/GOVERNANCE.md)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -55,7 +55,7 @@ SPIFFE is hosted by the [Cloud Native Computing Foundation](https://cncf.io) (CN
 * [CONTRIBUTING](/CONTRIBUTING.md)
 * [GOVERNANCE](/GOVERNANCE.md)
 
-### SIGs & Working Groups<a name="sigs"></a>
+### SIGs & Working Groups
```
spikecurtis: I'm guessing the anchor was in place to allow urls with a fragment `#sigs`. Since those URLs might still be in the wild, I don't see a reason to break them.

spikecurtis (dismissed):

README.md
evan2645: Fixed

PR #149, opened 2021-02-08
Fix Markdown syntax for Appendix A link

rturner3:
One of the links to Appendix A of the SPIFFE Workload Endpoint specification has an extra set of parentheses that get included into the link itself, resulting in a 404 when clicking the link.


justinburke:
Thank you for fixing this!

justinburke (approved):

PR #150, opened 2021-02-09
Cleanup vertical formatting of SPIFFE-ID.md table of contents

rturner3:
The other standards Markdown files use vertical formatting for the table of contents, which is easier to read than when all the items are on a single line. Add a couple spaces at the end of each table of contents line to force Markdown to insert newlines in-between each rendered table of contents item.


spikecurtis (approved):
LGTM

PR #152, opened 2021-02-18
Clarify & restrict trust domain names

spikecurtis:
Signed-off-by: Spike Curtis <spike@projectlucy.dev>

### DIFF
--- a/standards/SPIFFE-ID.md
+++ b/standards/SPIFFE-ID.md
@@ -45,7 +45,21 @@ The trust domain corresponds to the trust root of a system. A trust domain could
 
 Trust domain names are nominally self-registered, unlike public DNS there is no delegating authority that acts to assert and register a base domain name to an actual legal real-world entity, or assert that legal entity has fair and due rights to any particular trust domain name.
 
-The trust domain name is defined as the authority component of the URI - specifically, the `host` part of the authority. The `userinfo` and `port` parts of the authority component MUST NOT be set, and the `:` delimiter MUST NOT be present. Please see section 3.2 of [RFC 3986](https://tools.ietf.org/html/rfc3986) for more information.
+The trust domain name is defined as the authority component of the URI with the following restrictions applied:
+
+  * The `userinfo` and `port` parts of the authority component MUST NOT be set.
+  * The `host` part of the authority MUST be set to a non-zero-length string.
+  * The `host` part of the authority MUST be a registered name as defined by [section 3.2.2 of RFC 3986](https://tools.ietf.org/html/rfc3986#section-3.2.2).
+
+A registered name is defined as 
+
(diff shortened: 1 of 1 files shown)


standards/SPIFFE-ID.md
```diff
@@ -45,7 +45,21 @@ The trust domain corresponds to the trust root of a system. A trust domain could
...
+The trust domain name is defined as the authority component of the URI with the following restrictions applied:
+
+  * The `userinfo` and `port` parts of the authority component MUST NOT be set.
+  * The `host` part of the authority MUST be set to a non-zero-length string.
+  * The `host` part of the authority MUST be a registered name as defined by [section 3.2.2 of RFC 3986](https://tools.ietf.org/html/rfc3986#section-3.2.2).
+
+A registered name is defined as 
+
+```
+reg-name    = *( unreserved / pct-encoded / sub-delims )
```
justinburke, line 57: RFC 3986 defines `unreserved` as:
unreserved  = ALPHA / DIGIT / "-" / "." / "_" / "~"

A trust domain name of "~.~." or "____-____", for example, might be surprising. What do you think about providing a definition of unreserved and calling out that we do intend for those characters to be included as part of valid trust domain names?

RFC 3986 also defines `sub-delims`:
sub-delims  = "!" / "$" / "&" / "'" / "(" / ")" / "*" / "+" / "," / ";" / "="

The inclusion of `sub-delims` as part of `reg-name` is rather surprising to me. Do we really want to permit those characters? And if yes, then why are we excluding `IP-literal`, which includes square brackets?

Would it be simpler to define acceptable SPIFFE trust domain names as:
trustdomainname = 1*( alphasep / pct-encoded )
alphasep = ALPHA / DIGIT / "-" / "." / "_" 
pct-encoded = "%" HEXDIG HEXDIG

spikecurtis: Following your suggestion will stop me from using `;-)` as my trust domain name

spikecurtis: > then why are we excluding IP-literal, which includes square brackets?

My goal was to make validation easy.  `IP-literal` significantly ups the ante because it no longer is just loop over the characters and ensure they are in the allowed list.

justinburke: Per last week's SIG-Spec conversation, could we update this PR to define a valid host part to match the following character set: [a-z0-9.-]
The host part may not contain uppercase alphabetic characters, only lowercase is permitted.
Host part does not need to be a valid DNS name; for example, "-.-.-" is a valid trust domain name since it is non-empty and matches the above character set.

evan2645:
Superseded by https://github.com/spiffe/spiffe/pull/183

PR #155, opened 2021-03-10
Add SPIFFE Federation spec

evan2645:
This PR introduces the SPIFFE Federation specification.

The working version of this specification can be found here: https://docs.google.com/document/d/1ARs9dplNpw8MBtWwfVp4yInW9lK_fIYxGbP42Wuxeus


standards/SPIFFE_Federation.md
```diff
@@ -0,0 +1,316 @@
...
+
+While this fact may appear self evident, there are situations in which it may be less obvious. For example, in the case of a non-self-serving bundle endpoint wherein trust domain A serves a bundle for trust domain B, trust domain B is implicitly trusting trust domain A and its administrators to serve the correct bundle contents. Similarly, if serving a bundle from a hosting platform such as AWS S3, operators of the trust domain that the bundle in question represents are implicitly trusting AWS to serve the correct bundle contents.
+
+When choosing a location from which a SPIFFE bundle will be served, it is important to consider the trustworthiness of the parties involved.
+
+### 7.5. Authenticity of the Bundle Endpoint
+Ensuring the authenticity of the bundle endpoint is of paramount importance. This cannot be stated strongly enough. This section explores a number of considerations that should be made in ensuring the authenticity of bundle endpoints.
+
+#### 7.5.1. Bundle Endpoint URL Redirection
```
evan2645, line 276: The text in this section is an accepted suggestion from @spikecurtis. I have personally reviewed it and think it looks good, but it would benefit from further review by @justinburke and @azdagron

standards/SPIFFE_Federation.md
```diff
@@ -0,0 +1,316 @@
...
+
+Please see the [Security Considerations](#7-security-considerations) section for more information about the use of public certificate authorities.
+
+##### 5.2.1.1. Endpoint URL Requirements
+Bundle endpoint URLs utilizing `https_web` MUST have the scheme set to `https` and MUST NOT include userinfo in the authority component. Other components of the URL (as defined by [RFC 3986 section 3][7]) are not constrained by this specification.
+
+For example, the URL `https://host.example.com/trust_domain` is a valid SPIFFE bundle endpoint URL for the `https_web` profile type.
+
+##### 5.2.1.2. Endpoint Parameters
+The `https_web` profile MUST NOT require any additional parameters beyond those required for every profile (namely the trust domain name, profile type, and endpoint URL) in order to function.
```
sorindumitru, line 128: I think this should also include a parameter for CA certificate bundle for the endpoint. There's an implied default value that works in most cases, the system certificate pool, but it's not mandated that the certificate used for the endpoint needs to be issued by a public certificate authority(line 127 has that as a SHOULD, not a MUST).

evan2645: The canonical way to trust a private CA that works with RFC 6125 style validation (i.e. Web PKI) is to add the private root to the system store. If you're trying to accomplish CA pinning, my interpretation of this text is that it is still allowed so long as the implementation does not require it in order to work.

I think this brings up a related point though about `https_spiffe`... we should probably clarify that the parameter is optionally supplied but _not_ optionally supported.

sorindumitru: If it's still allowed to pin a CA, that should be good enough for me.

About adding the CA to the system store, i'm not sure that's the best approach for this. That ends up trusting the CA for anything on the system. Even for spire-server, it means that it's less secure to have multiple federation relationships that have different private CAs. So it might still be worth pointing out the implicit parameter here and making it optionally configurable.

spikecurtis: When we were designing the different profiles, I had assumed that people who wanted to use private CAs and limit their scope would choose `https_spiffe`.  In that profile, the endpoint is identified by a specific machine identity and authenticated by a specific cryptographic authority.

To the extent we thought people would use private CAs, I had assumed it would be in a large enterprise environment where the enterprise maintains its own CA, and all machines get the CA cert installed at provisioning time.

The use case you're describing seems to be about pinning specific bundle endpoints to specific (non-SPIFFE) CAs, which isn't something I thought much about as we were drafting this document. @sorindumitru can you talk a little bit more about your use case in the comments.  Why do you want to do it that way instead of using `https_spiffe`?

standards/SPIFFE_Federation.md
```diff
@@ -0,0 +1,316 @@
...
+
+SPIFFE is decentralized by nature. Each trust domain acts in its own capacity, under its own authority, and is administratively isolated from systems residing in other trust domains. Although trust domains delineate administrative and/or security domains, a core SPIFFE use case is enabling communication across these same boundaries where needed. Therefore, it is necessary to define a mechanism by which an entity may be introduced to a foreign trust domain, allowing it to authenticate identities issued by “other” SPIFFE authorities and allowing workloads in one trust domain to securely authenticate workloads in a foreign trust domain.
+
+A [SPIFFE bundle][2] is a self-contained resource that contains the public key material needed to authenticate identities in a particular trust domain. This document introduces a specification by which SPIFFE bundles can be securely fetched for the purpose of authenticating identities issued by external authorities. Included is information on how to serve a SPIFFE bundle, how to retrieve SPIFFE bundle, and how to authenticate the endpoints that serve them.
```
justinburke: I'm not entirely sure what is mean by "self-contained" here. The bundle doesn't, for example, include an identifier for a particular trust domain. Can we strike "self-contained"?

standards/SPIFFE_Federation.md
```diff
@@ -0,0 +1,316 @@
...
+
+## 2. Introduction
+Possession of a trust domain's bundle is required in order to validate SVIDs from it. Therefore, achieving SPIFFE federation necessitates the exchange of SPIFFE bundles between trust domains. This exchange occurs on a regular basis, allowing the contents of a trust domain's bundle to change over time.
```
justinburke: This exchange *should* occur on a regular basis
or
This exchange *is intended to* occur on a regular basis
or
This exchange *is designed to* occur on a regular basis

As currently stated, it read to me as a statement of fact that the exchange does occur on a regular basis.

This suggested edit will also help this text match that of Section 4.1 which specifies that clients SHOULD perform polling.

standards/SPIFFE_Federation.md
```diff
@@ -0,0 +1,316 @@
...
+
+Second, SPIFFE Federation is also used between trust domains residing in different companies or organizations. This case is similar to the first in that we are federating between SPIFFE deployments, however due to potential differences in implementation and administration, coordination is generally limited to the data exchanged in the SPIFFE Federation protocol described herein.
+
+Finally, SPIFFE Federation can also enable use cases for consumers that don't yet have a mature SPIFFE control plane deployed. For example, a hosted product may wish to authenticate its customers using their SPIFFE identities without having to internally implement or deploy SPIFFE. This can be accomplished by allowing a workload to directly fetch the customer's trust domain bundle in order to authenticate their callers, negating the need to commit to a full-blown SPIFFE deployment.
```
justinburke: obviating ?

standards/SPIFFE_Federation.md
```diff
@@ -0,0 +1,316 @@
...
+
+Deprecated keys SHOULD be removed from the trust bundle after a trust domain no longer has any active, valid SVIDs issued from those keys. Failure to follow these recommendations as keys are added to and removed from the bundle could result in transient cross-domain authentication failures. Requirements for updating the trust bundle do not apply to keys used to issue SVIDs for internal use only.
+
+Clients SHOULD periodically poll the endpoint for updates because the contents are expected to [change over time][4] - key validity periods on the order of weeks or even days is commonplace. Clients SHOULD poll at a frequency equal to the value of the bundle’s `spiffe_refresh_hint`, in seconds. If not set, a reasonably low default value should apply - five minutes is recommended.
+
+### 4.2. Managing Fetched Bundles
+Clients of the bundle endpoint MUST store the latest SPIFFE bundle each time it is retrieved. The sequence number field of the trust bundle SHOULD be used when comparing the freshness or ordering of two trust bundles. If the trust bundle omits the sequence number, operators SHOULD consider the most recently retrieved bundle to be up-to-date.
```
justinburke: MUST or SHOULD ?

What breaks if a client doesn't store the SPIFFE bundle?

standards/SPIFFE_Federation.md
```diff
@@ -0,0 +1,316 @@
...
+
+Bundle endpoint servers that support TLS-based profiles (e.g. `https_web` or `https_spiffe`) MUST adhere to the [Mozilla intermediate compatibility][5] requirements unless otherwise stated by the profile in use.
+
+### 5.1. Endpoint Parameters
+Prior to retrieving a bundle from a SPIFFE bundle endpoint, clients MUST be configured with the following three parameters: (1) the URL of the SPIFFE bundle endpoint, (2) the endpoint profile type, and (3) the trust domain name to associate with the bundle endpoint. The first two parameters indicate the location of the bundle endpoint and how to authenticate it. Since trust bundles do not contain a trust domain name, clients use the third parameter to associate a downloaded bundle with a specific trust domain name. Specific [endpoint profiles](#52-endpoint-profiles) (such as `https_spiffe`, as described below) MAY define additional mandatory configuration parameters.
```
justinburke, line 96: I feel like we should say something about what clients are supposed to do when a trust bundle contains unrecognized parameters:

Clients SHOULD ignore configuration parameters that are not defined in the endpoint profiles below.

standards/SPIFFE_Federation.md
justinburke: This is an interesting use case. And I, too, would like to hear more from @sorindumitru. Thanks for bringing this up!

So far, I think that we should certainly consider this use case, but shouldn't block this PR. Since the intent of `https_web` is to achieve a balance of security/paranoia with maximum compatibility (by "obviating the need for additional client configuration"), an additional parameter to specify a custom CA root cert may best belong in a new endpoint profile. (We'll also need to reconcile support here for a custom CA with redirect support. For example, if a custom CA parameter is specified but the endpoint returns a redirect to an endpoint which chains up to a different, perhaps well known public CA, what is a client supposed to do? And would a custom CA parameter override any use of the system-wide cert store or supplement it?)

standards/SPIFFE_Federation.md
```diff
@@ -0,0 +1,316 @@
...
+
+##### 5.2.1.2. Endpoint Parameters
+The `https_web` profile MUST NOT require any additional parameters beyond those required for every profile (namely the trust domain name, profile type, and endpoint URL) in order to function.
+
+##### 5.2.1.3. Serving a Bundle Endpoint
+SPIFFE bundle endpoint servers supporting the `https_web` transport type utilize standard TLS-protected HTTP (i.e. HTTPS). The server certificate used SHOULD be issued by a public certificate authority (as defined by the membership list of the CA/Browser forum), and MUST include the DNS name or IP address of the endpoint as an X.509 Subject Alternative Name (or Common Name). 
+
+As an interoperability concern, servers MUST NOT require client authentication to access the bundle endpoint; this includes both transport layer (e.g. client certificates) and HTTP-layer (e.g. Authentication headers) authentication schemes.
+
+Upon receiving an HTTP GET request for the correct path, the bundle endpoint server MUST respond with the most up-to-date version of the SPIFFE bundle available. The exact path value may be chosen by the operator, and appears as part of the bundle endpoint URL.
```
justinburke: and appears as *the path in* the bundle endpoint URL   ?

standards/SPIFFE_Federation.md
```diff
@@ -0,0 +1,316 @@
...
+
+In addition to the endpoint parameters required of all profiles, the `https_spiffe` profile requires additional endpoint client parameters as described in Endpoint Parameters below.
+
+##### 5.2.2.1. Endpoint URL Requirements
+Bundle endpoint URLs utilizing `https_spiffe` MUST have the scheme set to `https` and MUST NOT include userinfo in the authority component.  Other components of the URL (as defined by [RFC 3986 section 3][7]) are not constrained by this specification. 
+
+For example, the URL `https://host.example.com/trust_domain` is a valid SPIFFE bundle endpoint URL for the `https_spiffe` profile type.
+
+##### 5.2.2.2. Endpoint Parameters
+Bundle endpoint clients using the `https_spiffe` profile MUST be configured with the SPIFFE ID of the bundle endpoint server as well as a secure method for obtaining the trust bundle of the endpoint server's trust domain. A self-serving bundle endpoint is one in which the bundle endpoint server’s SPIFFE ID resides in the same trust domain as the bundle being fetched. Configured bundle endpoints may or may not be self-serving.
```
justinburke: Can we bold "self-serving bundle endpoint" to help readers identify this as a definition of a new term? Or put it in quotes or... something...

standards/SPIFFE_Federation.md
```diff
@@ -0,0 +1,316 @@
...
+Each SPIFFE Federation relationship is configured with the following parameters at minimum:
+* Trust domain name
+* Endpoint URL
+* Endpoint Profile
+
+It is important that these three parameters are configured explicitly, the values cannot be securely inferred from each other.
+
+For example, one might be tempted to infer the SPIFFE trust domain name from the host portion of the Endpoint URL. Firstly, we emphasize that a SPIFFE trust domain name need not be a registered DNS name.  Secondly, inferring this as a default is dangerous because it would allow anyone who can get a file served from a particular DNS name to assert trust roots for the SPIFFE trust domain of the same name. 
+
+Imagine a web hosting company called `spiffyhost.com`, that allows a customer, Alice, to serve web content at URLs like `https://spiffyhost.com/alice/<filename>`, and further, that `spiffyhost.com` operates an API secured by SPIFFE Federation with the SPIFFE trust domain name `spiffyhost.com`.  Imagine Alice sets up SPIFFE Federation with Bob, who is also a customer of `spiffyhost.com`, and Alice chooses to serve her trust bundle from `https://spiffyhost.com/alice/spiffe-bundle`.
```
justinburke: Is this domain name owned by SPIFFE? It's a real domain name....

standards/SPIFFE_Federation.md
justinburke: :shipit:

justinburke (changes requested):

standards/SPIFFE_Federation.md
justinburke: Hmm... the GitHub UI isn't showing me the text that I highlighted when I first created my comment. I'm not sure if that's a limitation of the UI or user error... in any case, this suggestion is to replace "negating" with "obviating".

standards/SPIFFE_Federation.md
justinburke: To clarify, the suggestion here is to replace "MUST store the latest" with "SHOULD store the latest".

standards/SPIFFE_Federation.md
justinburke: or "as the path of the bundle endpoint URL" ?

standards/SPIFFE_Federation.md
sorindumitru: The use case I had in mind is for when the different trust domains can't talk to each other directly but are still reachable through various gateways/proxies so `https_spiffe` might not work. An example would be `prod.example.com` couldn't directly talk with `dev.example.com` to fetch the bundle, but `dev` could still reach `prod` through gateways.

In that case the domains could reach out to a third party HTTP service that is reachable from both places to get access to the trust bundles. This case doesn’t require multiple certificates but I still think it would be wrong to add that CA to system pool. If it’s in the system pool, the same CA would be used for all TLS connections, e.g. TLS database connections.

I’m definitely ok with not blocking the PR on this. As Evan said, the wording is `MUST NOT require` so as long as it’s an optional parameter, implementations are free to have this.

I’m not really convinced that this needs to be a different profile, it just makes the root of trust for the endpoint configurable, but I can understand that this might make it harder for domains to federate in the future if they start with a private CA.

standards/SPIFFE_Federation.md
spikecurtis:
```suggestion
A [SPIFFE bundle][2] is a resource that contains the public key material needed to authenticate identities in a particular trust domain. This document introduces a specification by which SPIFFE bundles can be securely fetched for the purpose of authenticating identities issued by external authorities. Included is information on how to serve a SPIFFE bundle, how to retrieve SPIFFE bundle, and how to authenticate the endpoints that serve them.
```

standards/SPIFFE_Federation.md
spikecurtis:
```suggestion
Possession of a trust domain's bundle is required in order to validate SVIDs from it. Therefore, achieving SPIFFE federation necessitates the exchange of SPIFFE bundles between trust domains. This exchange should occur on a regular basis, allowing the contents of a trust domain's bundle to change over time.
```

standards/SPIFFE_Federation.md
spikecurtis:
```suggestion
Finally, SPIFFE Federation can also enable use cases for consumers that don't yet have a mature SPIFFE control plane deployed. For example, a hosted product may wish to authenticate its customers using their SPIFFE identities without having to internally implement or deploy SPIFFE. This can be accomplished by allowing a workload to directly fetch the customer's trust domain bundle in order to authenticate their callers, obviating the need to commit to a full-blown SPIFFE deployment.
```

standards/SPIFFE_Federation.md
spikecurtis: I'm ok with SHOULD.

If the client doesn't store the bundle at all, then it will need to issue another GET each time it is needed.  There are probably some legit use cases for that, like a workload that needs to validate an SVID once a day.

If the client stores the bundle, but doesn't store the latest bundle, then it could fall behind and fail validation of SVIDs.

standards/SPIFFE_Federation.md
spikecurtis: These are client config parameters, not parameters in the trust bundle.

I don't think we can reasonably (or should) say that implementers can't add additional parameters.  As a trivial example, I think it would be reasonable for an implementer to provide a config parameter that sets or overrides the refresh interval for a specific endpoint.

standards/SPIFFE_Federation.md
spikecurtis, line 128: I'm now wondering if this requirement is too strict.  Imagine a client written in a fully-portable framework that doesn't have access to the operating system's root CA store (or being deployed on an operating system that doesn't have a root CA store at all), and thus requires the CA store to be specified as a mandatory parameter.  Would we be willing to call that implementation non-complying?

standards/SPIFFE_Federation.md
spikecurtis:
```suggestion
Upon receiving an HTTP GET request for the correct path, the bundle endpoint server MUST respond with the most up-to-date version of the SPIFFE bundle available. The choice of path is not constrained by this specification.
```

The sentence as originally written and @justinburke 's suggested corrections veer a little to far into explaining "how to URL" IMO.

standards/SPIFFE_Federation.md
spikecurtis:
```suggestion
Bundle endpoint clients using the `https_spiffe` profile MUST be configured with the SPIFFE ID of the bundle endpoint server as well as a secure method for obtaining the trust bundle of the endpoint server's trust domain. A **self-serving** bundle endpoint is one in which the bundle endpoint server’s SPIFFE ID resides in the same trust domain as the bundle being fetched. Configured bundle endpoints may or may not be self-serving.
```

standards/SPIFFE_Federation.md
spikecurtis: Pretty sure it isn't.  I think it was unregistered when I made up this example.  Should we rewrite with `example.com`?

standards/SPIFFE_Federation.md
```diff
@@ -0,0 +1,316 @@
...
+7.5.3. [Endpoint Parameters](#753-endpoint-parameters)  
+7.6. [Chaining Trust with `https_spiffe`](#76-chaining-trust-with-https_spiffe)  
+
+## 1. Background
+The SPIFFE specifications define the documents and interfaces necessary for establishing a platform-agnostic workload identity framework that is capable of bridging systems in different domains without the need for implementing identity translation or credential exchange logic. They define a “[trust domain][1]”, which serves as an identity namespace.
+
+SPIFFE is decentralized by nature. Each trust domain acts in its own capacity, under its own authority, and is administratively isolated from systems residing in other trust domains. Although trust domains delineate administrative and/or security domains, a core SPIFFE use case is enabling communication across these same boundaries where needed. Therefore, it is necessary to define a mechanism by which an entity may be introduced to a foreign trust domain, allowing it to authenticate identities issued by “other” SPIFFE authorities and allowing workloads in one trust domain to securely authenticate workloads in a foreign trust domain.
```
matthewstevenson88: Should it say "asserted" rather than "issued"? Since an authority may issue a credential that asserts an identity, but it doesn't really issue the identity itself.

Similarly in the paragraph below.

Apologies if this is standard terminology in the SPIFFE specs, I'm new to reading this. :)

standards/SPIFFE_Federation.md
```diff
@@ -0,0 +1,316 @@
...
+
```
matthewstevenson88, line 47: In the last sentence, I think it should read "retrieve a" rather than "retrieve".

standards/SPIFFE_Federation.md
```diff
@@ -0,0 +1,316 @@
...
+A [SPIFFE bundle][2] is a self-contained resource that contains the public key material needed to authenticate identities in a particular trust domain. This document introduces a specification by which SPIFFE bundles can be securely fetched for the purpose of authenticating identities issued by external authorities. Included is information on how to serve a SPIFFE bundle, how to retrieve SPIFFE bundle, and how to authenticate the endpoints that serve them.
+
+## 2. Introduction
```
matthewstevenson88, line 50: This first paragraph has the first instance of the phrase "SPIFFE federation" in this doc. Is it worth defining exactly what this term means? Namely, calling out that "SPIFFE federation is the process of (cryptographically) authenticating to a peer in a different trust domain". 

I'm sure that this is spelled out in other SPIFFE docs, but it may be good to emphasize since "federation" can often be an ambiguous term.

standards/SPIFFE_Federation.md
```diff
@@ -0,0 +1,316 @@
...
+
+## 4. SPIFFE Bundle Endpoint
+A SPIFFE bundle endpoint is a resource (represented by a URL) that serves a copy of a SPIFFE bundle for a trust domain. SPIFFE control planes may both expose and consume these endpoints in order to transfer bundles between themselves, thereby achieving federation.
+
+The semantics of the SPIFFE bundle endpoint are similar to the `jwks_uri` mechanism defined in the OpenID Connect specification in that the bundle endpoint contains one or more public cryptographic keys used by a trust domain (for certifying identities within the trust domain). The bundle endpoint is an HTTPS URL that responds to an HTTP GET with a SPIFFE bundle. For more information about SPIFFE bundles and how they are encoded, please see the [SPIFFE Trust Domain and Bundle][3] specification.
```
matthewstevenson88: In the first sentence, should it read "...the bundle contains..." rather than "...the bundle endpoint contains..."?

standards/SPIFFE_Federation.md
```diff
@@ -0,0 +1,316 @@
...
+
+### 4.3. Endpoint Address Stability
+Once foreign trust domains begin relying on a specific endpoint URL, it is a delicate and error-prone procedure to migrate all clients of the endpoint to a replacement endpoint URL. Therefore, the safest course of action is to prefer a stable endpoint URL.
+
+## 5. Serving and Consuming a SPIFFE Bundle Endpoint
+This specification defines two supported profiles for SPIFFE bundle endpoint servers, both of which are based on HTTPS. One relies on the use of Web PKI to authenticate the endpoint and the other leverages SPIFFE authentication. SPIFFE bundle endpoint clients MUST support both of these profiles while SPIFFE bundle endpoint servers MUST support at least one.
```
matthewstevenson88, line 91: Can we replace both "MUST"s in the last sentence with "SHOULD"s?

A SPIFFE deployment could reasonably use an authentication mechanism that isn't TLS-based to fetch a SPIFFE bundle. For this to occur, we'd also need to relax the condition that a SPIFFE bundle endpoint MUST be an HTTPS server and say instead that it SHOULD be an HTTPS server.

matthewstevenson88: cc: @justinburke

standards/SPIFFE_Federation.md
```diff
@@ -0,0 +1,316 @@
...
+
+After establishing a TLS connection to the bundle endpoint and authenticating the presented server certificate, clients issue an HTTP GET for the path specified by the endpoint URL. The body of the response is a SPIFFE bundle. Clients MUST know the name of the trust domain that the endpoint URL represents prior to retrieving the trust bundle, ideally via explicit configuration; please see the [Security Considerations](#7-security-considerations) section for more information.
+
+Bundle endpoint servers MAY respond with HTTP redirect (as defined by [RFC 7231 section 6.4][8]).  Bundle endpoint clients SHOULD follow the redirect if the URL meets all the requirements of a valid bundle endpoint URL.  When connecting to the new URL, the same TLS considerations MUST be applied as connecting to the original URL.  Bundle endpoint clients SHOULD use the configured endpoint URL for each bundle refresh and SHOULD NOT permanently store the location for future fetches. See the [Security Considerations](#7-security-considerations) for more information. 
+
+#### 5.2.2. SPIFFE Authentication (`https_spiffe`)
```
matthewstevenson88, line 146: Do we gain anything by requiring that the bundle endpoint's cert be itself an X509 SVID? You could imagine that a client gets a private PKI root cert for the bundle endpoint in an out-of-band way and adds it to their root store. Or would that fall under the `https_web` profile?

standards/SPIFFE_Federation.md
```diff
@@ -0,0 +1,316 @@
...
+
+Once the connection has been successfully established, and a copy of the bundle has been received, it is stored along with the name of the trust domain it belongs to. The contents of the bundle (e.g. CA certificates, JWT signing keys, etc) can now be distributed for the purposes of validating SVIDs originating from the foreign trust domain.
+
+The exact nature and mechanism by which this distribution occurs is an implementation detail, and is out of scope for this document. For more information on how SPIFFE-aware workloads can receive bundle updates, please see the [SPIFFE Workload API][10] specification.
+
+### 6.2. Maintaining a Relationship
+SPIFFE bundle endpoint clients poll the bundle endpoint periodically for updates. When an update is detected, the stored bundle representing the endpoint's foreign trust domain is updated to match. The updated content is then distributed so that validators can add new keys and drop revoked keys as necessary. Again, the exact method(s) for distributing this update to validators is out of scope for this document.
```
matthewstevenson88: Per another comment in this PR, I think it should be: "...clients SHOULD poll..."

standards/SPIFFE_Federation.md
```diff
@@ -0,0 +1,316 @@
...
+
+Imagine a web hosting company called `spiffyhost.com`, that allows a customer, Alice, to serve web content at URLs like `https://spiffyhost.com/alice/<filename>`, and further, that `spiffyhost.com` operates an API secured by SPIFFE Federation with the SPIFFE trust domain name `spiffyhost.com`.  Imagine Alice sets up SPIFFE Federation with Bob, who is also a customer of `spiffyhost.com`, and Alice chooses to serve her trust bundle from `https://spiffyhost.com/alice/spiffe-bundle`.
+
+![A diagram illustrating the relationships between Alice, Bob, and `spiffyhost.com`](https://raw.githubusercontent.com/evan2645/spiffe/2c6d5bd6c9b7e8aafc01ec577e7be53242e18e06/standards/img/spiffe_federation_spiffyhost_example.png)  
+*Figure 5: A diagram illustrating the relationships between Alice, Bob, and `spiffyhost.com`.*
+
+If Bob’s control plane implicitly gets the trust domain name from the URL, this would then allow Alice to impersonate the trust domain spiffyhost.com!
```
matthewstevenson88: nit: "spiffyhost.com" should be written `spiffyhost.com` as it is in the previous paragraph.

standards/SPIFFE_Federation.md
evan2645: +1 for SHOULD

Spike's example of validating an SVID once a day is a good one (cron jobs come to mind). I think this could also be important for safety. For example, if a client notices that the latest SPIFFE bundle has no keys present, or some similarly suspect change, they may choose to ignore it.

standards/SPIFFE_Federation.md
evan2645: I think it's a good idea to stick with the reserved names. All example subdomains are also reserved, so we could do something like `host-my-page.example.com`. Choosing a name that carries with it the notion that folks may self-serve content there will help drive the example home.

standards/SPIFFE_Federation.md
evan2645: I think that in your example, the software as a whole would be configured with a path to the CA store (as opposed to being defined on a per-endpoint basis)

standards/SPIFFE_Federation.md
spikecurtis: > A SPIFFE deployment could reasonably use an authentication mechanism that isn't TLS-based to fetch a SPIFFE bundle.

I don't disagree, but this spec as written is based on TLS / HTTPS.  A different authentication mechanism would be a new profile, distinct from the ones defined in the spec.  We want it to be the case that if the client software complies with the spec and the server software complies with the spec, then it will work.  If we relax this to SHOULD, then you could have server software that doesn't use TLS and a client that insists on it, and interop will fail.

standards/SPIFFE_Federation.md
spikecurtis: That's true.

More generally where my mind is going is that we've made an assumption that given the endpoint URL, the client will always have all the information it needs to access a particular endpoint server and validate it.  Does that assumption hold in all cases?  Let's think networking: could an implementation insist that you configure the network interface, or resolver, or IP version, or... ?  For example, software that gets deployed on a gateway between two networks with different security classifications---you might insist that each endpoint gets marked with the appropriate network to prevent DNS queries of secret names being executed on unclass networks.

Maybe such a scenario is too exotic to worry about; WDYT?

standards/SPIFFE_Federation.md
spikecurtis:
```suggestion
SPIFFE is decentralized by nature. Each trust domain acts in its own capacity, under its own authority, and is administratively isolated from systems residing in other trust domains. Although trust domains delineate administrative and/or security domains, a core SPIFFE use case is enabling communication across these same boundaries where needed. Therefore, it is necessary to define a mechanism by which an entity may be introduced to a foreign trust domain, allowing it to authenticate credentials issued by “other” SPIFFE authorities and allowing workloads in one trust domain to securely authenticate workloads in a foreign trust domain.
```

standards/SPIFFE_Federation.md
spikecurtis: Github attaches this to lines above, but it refers to lines below, and I agree.

standards/SPIFFE_Federation.md
spikecurtis: I feel like our use of "federation" is fairly disambiguated by the Background section.

standards/SPIFFE_Federation.md
spikecurtis: I'm assuming here that in your scenario the client checks the domain name from the URI against the Common Name or DNS Subject Alt Name on the cert. In which case, yes, this would be `https_web`.

In `https_spiffe` the client checks the configured SPIFFE ID against the URI Subject Alt Name.

standards/SPIFFE_Federation.md
evan2645: Hmm... that is a good question...

IMO, a big part of this requirement is interop as it relates to the advertisement of endpoint parameters. A custom implementation may require such details (as illustrated in your most recent example), but it's kind of bespoke at that point? Can you say that such an implementation is conformant? With the current text, the answer is no... but does that person care? They'll certainly tell others that they are conformant!

We've had at least one community request for some rigidity around the way that federation relationships are configured such that it is easy to distribute coordinates or parameters. While codifying an actual format by which this occurs feels out-of-scope for this PR, it _does_ feel valuable to align on what the contents of such a document will be. So maybe the real question here is, do we want to try to preserve this property or not?

standards/SPIFFE_Federation.md
evan2645: It does seem, to me, beneficial to have a "SPIFFE Federation is ..." statement right around here. I agree that we have clearly articulated the picture, but a blunt statement could never hurt?

The suggestion tool doesn't work with this comment, here is what I'd propose:
```
## 2. Introduction
SPIFFE Federation is the act of obtaining the materials necessary to validate SPIFFE identities in a different trust domain, and providing said materials to the workloads performing the authentication. ...
```

spikecurtis: "SPIFFE Federation is the act of obtaining the materials necessary to authenticate identity credentials (SVIDs) issued by a different trust domain, and providing said materials to the workloads performing the authentication."

I like "authenticate credentials" better than "validate identities", which might be confused with things like validating a purported SPIFFE ID is valid according to our new rules about character set.

standards/SPIFFE_Federation.md
evan2645: > If we relax this to SHOULD, then you could have server software that doesn't use TLS and a client that insists on it, and interop will fail.

+1. The spec is about interop, and conformant software should work together. I don't think that means you can't write software that doesn't conform! Of course, you are free to write the software of your choosing ... however, if you write software that only supports an unspecified profile, then the spec has not accomplished much and you cannot reliably advertise conformance to your consumers and expect that to work. For example, even in a big corporation, if you say that you have a conformant SPIFFE federation thing but it uses a custom profile, clients or customers can't pick up off-the-shelf software and expect that to work.

matthewstevenson88: Thanks both for your replies. That's totally fair - I understand that you want any client/server pair using software that satisfies the spec to be able to fetch each other's trust bundle. 

I was coming at this question from the angle of: instead of having the control plane or trusted third party push the endpoint parameters, just have it push the SPIFFE bundle instead, and see if reading the bundle from disk could be made into a new endpoint profile. I'll file an issue to discuss further, since I don't want to clutter the PR. :)

standards/SPIFFE_Federation.md
justinburke: Here's my suggested edit. I think it would be clearer to first describe the abstract idea of federation followed by an overview of the mechanics.

"SPIFFE Federation enables cross-trust domain authentication of identity credentials (SVIDs). An operator first acquires the materials necessary for authenticating identity credentials of a remote trust domain and then distributes the materials to local workloads; workloads configured with the remote trust domain authentication materials may then authenticate SVIDs issued by the remote trust domain."

standards/SPIFFE_Federation.md
```diff
@@ -0,0 +1,316 @@
...
+
+## 2. Introduction
+Possession of a trust domain's bundle is required in order to validate SVIDs from it. Therefore, achieving SPIFFE federation necessitates the exchange of SPIFFE bundles between trust domains. This exchange occurs on a regular basis, allowing the contents of a trust domain's bundle to change over time.
+
+To achieve this, SPIFFE federation defines a "bundle endpoint", which is a URL that serves up a SPIFFE bundle for a particular trust domain. Also defined are a set of "endpoint profiles", which specify the protocol and authentication semantics used between bundle endpoint servers and clients. Finally, this document further specifies the behavior of bundle endpoint clients and servers, and the management of federation relationships and the resulting bundle data.
```
justinburke: s/SPIFFE federation/SPIFFE Federation/ for consistency.

```suggestion
To achieve this, SPIFFE Federation defines a "bundle endpoint", which is a URL that serves up a SPIFFE bundle for a particular trust domain. Also defined are a set of "endpoint profiles", which specify the protocol and authentication semantics used between bundle endpoint servers and clients. Finally, this document further specifies the behavior of bundle endpoint clients and servers, and the management of federation relationships and the resulting bundle data.
```

standards/SPIFFE_Federation.md
justinburke:
```suggestion
The semantics of the SPIFFE bundle endpoint are similar to the `jwks_uri` mechanism defined in the OpenID Connect specification in that the bundle contains one or more public cryptographic keys used by a trust domain (for certifying identities within the trust domain). The bundle endpoint is an HTTPS URL that responds to an HTTP GET with a SPIFFE bundle. For more information about SPIFFE bundles and how they are encoded, please see the [SPIFFE Trust Domain and Bundle][3] specification.
```

standards/SPIFFE_Federation.md
justinburke:
```suggestion
Clients of the bundle endpoint SHOULD store the latest SPIFFE bundle each time it is retrieved. The sequence number field of the trust bundle SHOULD be used when comparing the freshness or ordering of two trust bundles. If the trust bundle omits the sequence number, operators SHOULD consider the most recently retrieved bundle to be up-to-date.
```

standards/SPIFFE_Federation.md
justinburke: @matthewstevenson88, thanks for the suggestion. This is a topic that has come up in SIG-Spec discussions in the past. IIRC, the outcome has been along the lines of what both Spike and Evan have mentioned above: the focus of the spec is interoperability. If an operator wants to create an internal use implementation that, for example, pushes bundle files to disk, then that adaption would be okay since it's for internal use.
It's also possible that we could discuss standardizing that interface, but it'd be a new profile. Totally fine to consider additional profiles!

standards/SPIFFE_Federation.md
justinburke: Seems that I had previously misread this section. Thanks for pointing that out.

standards/SPIFFE_Federation.md
justinburke: Accommodating the situation above (needing to specify root cert store or networking configuration), seems to me to be a different layer of the software stack. We assert here as a prerequisite that the client has been configured with a certificate store. I'm in favor of leaving the text as is.

standards/SPIFFE_Federation.md
justinburke:
```suggestion
SPIFFE bundle endpoint clients SHOULD poll the bundle endpoint periodically for updates. When an update is detected, the stored bundle representing the endpoint's foreign trust domain is updated to match. The updated content is then distributed so that validators can add new keys and drop revoked keys as necessary. Again, the exact method(s) for distributing this update to validators is out of scope for this document.
```

standards/SPIFFE_Federation.md
justinburke: This search/replace needs to be performed throughout this file.

standards/SPIFFE_Federation.md
```diff
@@ -0,0 +1,316 @@
...
+
+### 7.2. Explicitly-defined Endpoint Parameters
+Each SPIFFE Federation relationship is configured with the following parameters at minimum:
+* Trust domain name
+* Endpoint URL
+* Endpoint Profile
+
+It is important that these three parameters are configured explicitly, the values cannot be securely inferred from each other.
+
+For example, one might be tempted to infer the SPIFFE trust domain name from the host portion of the Endpoint URL. Firstly, we emphasize that a SPIFFE trust domain name need not be a registered DNS name.  Secondly, inferring this as a default is dangerous because it would allow anyone who can get a file served from a particular DNS name to assert trust roots for the SPIFFE trust domain of the same name.
```
evan2645:
```suggestion
For example, one might be tempted to infer the SPIFFE trust domain name from the host portion of the Endpoint URL. Firstly, we emphasize that a SPIFFE trust domain name need not be a registered DNS name.  Secondly, inferring this as a default is dangerous because it could allow anyone who can get a file served from a particular DNS name to assert trust roots for the SPIFFE trust domain of the same name. 
```

standards/SPIFFE_Federation.md
```diff
@@ -0,0 +1,316 @@
...
+Each SPIFFE Federation relationship is configured with the following parameters at minimum:
+* Trust domain name
+* Endpoint URL
+* Endpoint Profile
+
+It is important that these three parameters are configured explicitly, the values cannot be securely inferred from each other.
+
+For example, one might be tempted to infer the SPIFFE trust domain name from the host portion of the Endpoint URL. Firstly, we emphasize that a SPIFFE trust domain name need not be a registered DNS name.  Secondly, inferring this as a default is dangerous because it would allow anyone who can get a file served from a particular DNS name to assert trust roots for the SPIFFE trust domain of the same name. 
+
+Imagine a web hosting company called `spiffyhost.com`, that allows a customer, Alice, to serve web content at URLs like `https://spiffyhost.com/alice/<filename>`, and further, that `spiffyhost.com` operates an API secured by SPIFFE Federation with the SPIFFE trust domain name `spiffyhost.com`.  Imagine Alice sets up SPIFFE Federation with Bob, who is also a customer of `spiffyhost.com`, and Alice chooses to serve her trust bundle from `https://spiffyhost.com/alice/spiffe-bundle`.
```
evan2645:
```suggestion
For example, imagine a web hosting company called `spiffyhost.com` that allows a customer, Alice, to serve web content at URLs like `https://spiffyhost.com/alice/<filename>`, and further, that `spiffyhost.com` operates an API secured by SPIFFE Federation with the SPIFFE trust domain name `spiffyhost.com`.  Imagine Alice sets up SPIFFE Federation with Bob, who is also a customer of `spiffyhost.com`, and Alice chooses to serve her trust bundle from `https://spiffyhost.com/alice/spiffe-bundle`.
```

standards/SPIFFE_Federation.md
evan2645: The reference to registered DNS names should be moved to the end since it is not the focal point of this conversation

standards/SPIFFE_Federation.md
evan2645: From SIG-Spec call: we want both the concept and the actual definition captured here. Something like below

SPIFFE Federation enables cross-trust domain authentication of identity credentials (SVIDs). Specifically, it is the act of obtaining ... a bundle

justinburke (dismissed):

standards/SPIFFE_Federation.md
```diff
@@ -250,14 +250,14 @@ Imagine a web hosting company called `spiffyhost.com`, that allows a customer, A
 ![A diagram illustrating the relationships between Alice, Bob, and `spiffyhost.com`](https://raw.githubusercontent.com/evan2645/spiffe/2c6d5bd6c9b7e8aafc01ec577e7be53242e18e06/standards/img/spiffe_federation_spiffyhost_example.png)
```
justinburke: Is spiffyhost.com still in the current PR? (unclear via UI)

justinburke: Sorry, please ignore my last comment.

standards/SPIFFE_Federation.md
```diff
@@ -243,16 +245,16 @@ Each SPIFFE Federation relationship is configured with the following parameters
 
 It is important that these three parameters are configured explicitly, the values cannot be securely inferred from each other.
 
-For example, one might be tempted to infer the SPIFFE trust domain name from the host portion of the Endpoint URL. Firstly, we emphasize that a SPIFFE trust domain name need not be a registered DNS name.  Secondly, inferring this as a default is dangerous because it would allow anyone who can get a file served from a particular DNS name to assert trust roots for the SPIFFE trust domain of the same name. 
+For example, one might be tempted to infer the SPIFFE trust domain name from the host portion of the Endpoint URL. Inferring this as a default is dangerous because it could allow anyone who can get a file served from a particular DNS name to assert trust roots for the SPIFFE trust domain of the same name.
```
justinburke: s/. Inferring this as a default/; inferring a relationship/ ?

standards/SPIFFE_Federation.md
```diff
@@ -0,0 +1,318 @@
...
+
+### 4.1. Adding and Removing Keys
+Operators of a trust domain MAY introduce or remove keys used to issue SVIDs within the trust domain as needed (e.g. as part of an internal key rotation process). When adding new keys, an updated trust bundle containing the keys SHOULD be published at the bundle endpoint sufficiently in advance that foreign trust domains have an opportunity to retrieve and internally disseminate the new bundle contents; the recommended advance time is 3-5 times the bundle’s `spiffe_refresh_hint`. At a minimum, new keys MUST be published at the bundle endpoint prior to the keys being used to issue SVIDs. 
+
+Deprecated keys SHOULD be removed from the trust bundle after a trust domain no longer has any active, valid SVIDs issued from those keys. Failure to follow these recommendations as keys are added to and removed from the bundle could result in transient cross-domain authentication failures. Requirements for updating the trust bundle do not apply to keys used to issue SVIDs for internal use only.
```
spikecurtis:
```suggestion
Deprecated keys SHOULD be removed from the trust bundle after a trust domain no longer has any active, valid SVIDs issued from those keys. Failure to follow these recommendations as keys are added to and removed from the bundle could result in transient cross-domain authentication failures.

Requirements for updating the trust bundle do not apply to keys used to issue SVIDs for internal use only.
```

Placed at the end of the paragraph it is in might be interpreted as only applying to removing keys, but it applies to both adding and removing, so should be in its own paragraph.

spikecurtis (dismissed):
One small (whitespace-only) suggestion, but LGTM.

azdagron (dismissed):
This is looking great. :+1:

standards/SPIFFE_Federation.md
evan2645: From SIG-Spec call: This is dangerous because ...

justinburke (approved):

azdagron (approved):

spikecurtis (approved):

evan2645:
Thanks to everyone here for reviewing and leaving your feedback, it's very much appreciated!

Please direct any further feedback or questions to GitHub Issues.

PR #167, opened 2021-04-21
Nominate Jeyappragash JJ for 2021H1 SSC Election

evan2645:


evan2645:
JJ declined to participate in this election due to his existing time commitment with CNCF's SIG-Security, however he has volunteered to participate in the next election cycle (should he be nominated) scheduled to take place in October 2021. Thank you JJ for your consideration - we're looking forward to the next election cycle.

PR #174, opened 2021-04-28
Nominate Dan Feldman and Mariano Kunzu for 2021H1 SSC Election

ai2017:


evan2645:
Hello @ai2017 and thanks for the nominations! Unfortunately, nominations [closed about a week ago](https://github.com/spiffe/spiffe/issues/160#issuecomment-824598415). The polls are already open, so I'm afraid it's too late to add any more nominees.

The next election will occur in October. Thanks again for your participation, and sorry for the confusion.

PR #180, opened 2021-05-19
Fix links in README following main branch rename

evan2645:


azdagron (approved):
This reminds me that we need to do the same thing in SPIRE...

PR #191, opened 2021-07-15
Add SPIRE Contributor sync meeting to SIG-SPIRE description

rturner3:
The SPIRE Contributor sync meeting has been ongoing for several months
and is used as a forum for contributors to meet with SPIRE maintainers
to get feedback and direction on currently proposed code changes to the
project. Mentioning the existence of the meeting here to more clearly
advertise it to current and future contributors to SPIRE.

### DIFF
--- a/community/sig-spire/README.md
+++ b/community/sig-spire/README.md
@@ -3,22 +3,36 @@
 A Special Interest Group (SIG) concerning SPIRE, the official SPIFFE implementation.
 
 ### Meetings:
+* [Calendar ICS](https://calendar.google.com/calendar/ical/c_q2don6m2b33gljqftauib3hnuk%40group.calendar.google.com/public/basic.ics)
+
+#### SIG-SPIRE
 * Date/Time: Every other Thursday @ 10:30am Pacific
 * [Meetings Notes](https://docs.google.com/document/d/1IgpCkvSRSoY9Xd16gFQJJ1KP8sLZ7EE39cEjBK_UIg4)
-* [Calendar ICS](https://calendar.google.com/calendar/ical/c_q2don6m2b33gljqftauib3hnuk%40group.calendar.google.com/public/basic.ics)
+
+##### Goals:
+* Discuss SPIRE project direction and roadmap
+* Provide a high-bandwidth forum in which the community can voice needs and make proposals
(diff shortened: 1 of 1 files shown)


evan2645:
Thanks for this @rturner3 

@dfeldman we had originally opted to put the Contributor Sync calendar event on the SIG-SPIRE calendar since that felt like the right place for it, and a dedicated calendar for the contributor sync didn't really seem necessary. It's not really a SIG per se, but it is pretty related to SIG-SPIRE.

How do you feel about this update? One thing I noticed is that the `Lead` section is scoped to the whole doc rather than to specific meetings. The SIG-SPIRE calendar still feels like the right place for the event IMO, but I'm curious to hear how you'd like to present the information here.

dfeldman:
I'm open to any change here, no worries. Maybe separate out the contributor sync and SIG-SPIRE sections?

rturner3:
> I'm open to any change here, no worries. Maybe separate out the contributor sync and SIG-SPIRE sections?

Based on this idea, I gave it a little more thought. Throwing out a few options on the table about how we want to manage this meeting that we can consider:
1. Separate the meetings into different sections in the SIG-SPIRE doc (suggestion from @dfeldman). Since the meetings are currently on the same calendar, if we separated the two meetings into separate sections in the doc, folks would see the same set of both meetings when importing the calendar file from either section of the doc.

2. Same option as 1), but also split out the SPIRE Contributor sync into a separate calendar invite such that the SIG-SPIRE and Contributor Sync sections of the doc have different calendars. This would impact anyone who currently attends the SPIRE Contributor Sync, and we would need to send a message to the community to subscribe to the new invite. That could be done over Slack as well as in the next Contributor Sync meeting.

3. Classify the Contributor Sync as another meeting of SIG-SPIRE (current implicit approach + what's currently drafted in this PR).

I'm open to any of these. Which option sounds best to you, @evan2645 + @dfeldman?

dfeldman:
I would recommend keeping them on the same Google calendar at least, since it is confusing having too many calendar subscriptions.

evan2645 (approved):
We should just go ahead and take this since it's better than what we've got now :)

Sorry for the delay @rturner3

PR #206, opened 2021-11-16
Add maturity badge

evan2645:
This commit adds the SPIFFE maturity badge to this repo. Even though
there are many things here to which the maturity badge does not really
apply, I think it makes sense to display here as it is also where the
spec text currently lives.

I'm not in love with the formatting but wasn't able to come up with anything
better.


lumjjb (approved):
Love it! This is awesome!

PR #212, opened 2022-02-15
Adjust timings for election process, add clarity

amoore877:
- Election process had an ambiguous "T-X" format for timing, but no clarity on what "T" is. Chose this to be "Term Start"
- Extended cycle by one week, so allow for final formalities and confirmations once polls are closed.
- Remove redundant information elsewhere.


ajessup (approved):
LGTM!

evan2645 (approved):
🤗 thanks @amoore877!

PR #215, opened 2022-03-18
Add new security considerations section for SPIFFE IDs

evan2645:
The SPIFFE ID spec has within it a set of character restrictions that
are intended to mitigate confusion and potential exploitation around
how to interpret an ID and the equality of IDs. It is important to
communicate the reasons and necessity of these restrictions to our users
so that they are preserved in their implementations.

This commit adds a new security considerations section that details why
these restrictions are in place, and touches up a few spots that have
fallen out of date following the changes in these restrictions. It is a
follow-on to the change made in PR #183.

### DIFF
--- a/standards/SPIFFE-ID.md
+++ b/standards/SPIFFE-ID.md
@@ -18,6 +18,8 @@ For more general information about SPIFFE, please see the [Secure Production Ide
 2.1.1. [Trust Domain Name Collisions](#211-trust-domain-name-collisions)  
 2.2. [Path](#22-path)  
 2.3. [Maximum SPIFFE ID Length](#23-maximum-spiffe-id-length)  
+2.4. [SPIFFE ID Parsing](#24-spiffe-id-parsing)  
+2.5. [SPIFFE ID Equivalency](#24-spiffe-id-equivalency)  
 3\. [SPIFFE Verifiable Identity Document](#3-spiffe-verifiable-identity-document)  
 3.1. [SVID Trust](#31-svid-trust)  
 3.2. [SVID Components](#32-svid-components)  
@@ -28,6 +30,8 @@ For more general information about SPIFFE, please see the [Secure Production Ide
 4.1.2. [Scope and Influence](#412-scope-and-influence)  
 4.1.3. [Interpretation](#413-interpretation)  
 4.1.4. [Veracity](#414-veracity)  
+4.2. [ID Equivalency](#42-id-equivalency)  
+5\. [Appendix A](#5-appendix-a)  
 
 
 ## 1. Introduction
@@ -109,11 +113,17 @@ Paths MAY be hierarchical - similar to filesystem paths. The specific meaning of
 
 URIs, as defined by [RFC 3986](https://tools.ietf.org/html/rfc3986), do not have a maximal length. As an interoperability consideration, SPIFFE implementations MUST support SPIFFE URIs up to 2048 bytes in length and SHOULD NOT generate URIs of length greater than 2048 bytes. [RFC 3986](https://tools.ietf.org/html/rfc3986) permits only ASCII characters, thus the recommended maximum length of a SPIFFE ID is 2048 bytes.
 
-All URI components contribute to the URI length, including the "spiffe" scheme, "://" separator, trust domain name, and path component. Non-ASCII characters contribute to the URI length after they are percent encoded as ASCII characters. Note that [RFC 3986](https://tools.ietf.org/html/rfc3986) defines a maximum length of 255 characters for the "host" component of a URI; therefore a maximum length of a trust domain name is 255 bytes.
+All URI components contribute to the URI length, including the "spiffe" scheme, "://" separator, trust domain name, and path component. Note that [RFC 3986](https://tools.ietf.org/html/rfc3986) defines a maximum length of 255 characters for the "host" component of a URI; therefore a maximum length of a trust domain name is 255 bytes.
 
 ### 2.4. SPIFFE ID Parsing
 
 SPIFFE IDs follow the URI specification as defined by [RFC 3986](https://tools.ietf.org/html/rfc3986). The scheme and trust domain name of the SPIFFE ID are case-insensitive. The path is case-sensitive.
+
+### 2.5. SPIFFE ID Equivalency
+
+Two SPIFFE IDs are equivalent if and only if they match on a byte-for-byte basis. Note that since SPIFFE IDs allow only ASCII characters without percent-encoding, and SPIFFE also forbids capital letters in the `host` part of the authority (which is traditionally case-insensitive), comparisons of legal SPIFFE IDs are disambiguated.
+
+It is important to try to minimize the amount of processing done on SPIFFE IDs received from untrusted sources prior to comparing them, as it is possible for some URI parsers to normalize an illegal SPIFFE ID into a legal one. Please see Security Considerations [section 4.2](#42-id-equivalency) for more information.
 
 ## 3. SPIFFE Verifiable Identity Document
 A SPIFFE Verifiable Identity Document (SVID) is the mechanism through which a workload communicates its identity to a resource or caller. An SVID is considered valid if it has been signed by an authority within the SPIFFE ID's trust domain.
@@ -184,4 +194,30 @@ Operators should carefully consider, on a case by case basis, whether or not a g
 For example, imagine that the operator of trust domain A and trust domain B have a shared customer and a business relationship. During normal operation in trust domain A, a specialized process authenticates the customer and the trust domain’s authority creates an SVID that embeds an assertion indicating the customer’s identity. The purpose of this process is to mitigate unauthorized access to customer data in the event that an intermediary service within the trust domain is compromised by proving that the customer was authenticated and did in fact make this request.
 
 Now imagine that a request for the customer’s data is received by the storage service in trust domain A, except the caller presented an SVID from trust domain B. Even though the presented SVID may have the necessary assertion indicating that the shared customer was authenticated and authorized the request, it is a bad idea to blindly assume that trust domain B has indeed authenticated your customer. If trust domain B is compromised, or if it has a malicious internal actor, it could claim to have authenticated any user it wishes, thus creating the very circumstance that the measure was designed to mitigate in the first place.
+
+### 4.2. ID Equivalency
+The comparison of SPIFFE IDs is a security critical operation. In allowing for internationalization, as well as compatibility with the DNS system, the URI standard which SPIFFE IDs rely upon has made a handful of decisions which complicate the equivalency process. Specifically, case-insensitivity, ambiguous percent-encoding rules and support for UTF-8, and the implementation of many conditionally-special characters can make the topic a confusing one.
+
+This specification has specifically forbidden the use of characters and encoding schemes that complicate comparison (please see sections [2.1](#21-trust-domain) and [2.2](#22-path) for more information). Thus, any legal SPIFFE ID can be considered safe. It will not be misinterpreted or transformed by traditional URI libraries, and it can be safely compared with other legal IDs on a byte-for-byte basis.
+
+It is still possible for a legal URI (but illegal SPIFFE ID) to be legally processed according to URI normalization rules _into_ a legal SPIFFE ID. Thus, under certain circumstances, it is possible for an illegal SPIFFE ID to be passed through a URI parser and produce a legal SPIFFE ID. These two SPIFFE IDs **will not match** on a byte-for-byte basis, however there exists room for confusion.
+
+This is an important detail to keep in mind, particularly when accepting untrusted input that is not signed over by a trust domain authority (e.g. user input from a web app). Untrusted input should always be validated against the SPIFFE ID ruleset prior to accepting and parsing with a traditional URI parser. Please see [Appendix A](#5-appendix-a) for a brief example.
+
+## 5. Appendix A. Lightweight SPIFFE ID validation
+SPIFFE IDs have a strict character set that is designed to be as consistent and as easy to validate as possible. Here is a logic example demonstrating a simple SPIFFE ID validation mechanism:
+1. Total number of characters is less than 2048
+1. Starts with "spiffe://"
+1. Strip prefix "spiffe://"
+1. The (new) first character must not be a `/`
+1. From the beginning, strip one character at a time:
+    1. Ensure it is either a lowercase alpha-numeric, a `.`, a `-`, or a `/`
+    1. When `/` is detected, stop
+    1. The number of stripped characters should not exceed 255, not including `/`
+1. What remains is the path - scan it for the following conditions in sequential order:
+    1. The string does not start or end with `/`
+    1. Includes only uppercase or lowercase alpha-numeric characters, `.`s, `-`s, `_`s, and `/`s
+    1. The character sequences `//`, `/./`, and `/../` do not appear anywhere in the string
+
+If the above checks are successful, the SPIFFE ID is valid.
 


standards/SPIFFE-ID.md
```diff
@@ -185,3 +195,29 @@ For example, imagine that the operator of trust domain A and trust domain B have
...
+1. From the beginning, strip one character at a time:
+    1. Ensure it is either a lowercase alpha-numeric, a `.`, a `-`, or a `/`
+    1. When `/` is detected, stop
+    1. The number of stripped characters should not exceed 255, not including `/`
+1. What remains is the path - scan it for the following conditions in sequential order:
+    1. The string does not start or end with `/`
+    1. Includes only uppercase or lowercase alpha-numeric characters, `.`s, `-`s, `_`s, and `/`s
+    1. The character sequences `//`, `/./`, and `/../` do not appear anywhere in the string
+
+If the above checks are successful, the SPIFFE ID is valid.
```
evan2645, line 222: I think @spikecurtis's python script can be linked to from here. Maybe we can add it to this PR and check it into a new `/standards/examples` directory?

evan2645: Alternatively, we could link to it directly from section 4.2 and omit this appendix item

standards/SPIFFE-ID.md
```diff
@@ -109,12 +113,18 @@ Paths MAY be hierarchical - similar to filesystem paths. The specific meaning of
...
 URIs, as defined by [RFC 3986](https://tools.ietf.org/html/rfc3986), do not have a maximal length. As an interoperability consideration, SPIFFE implementations MUST support SPIFFE URIs up to 2048 bytes in length and SHOULD NOT generate URIs of length greater than 2048 bytes. [RFC 3986](https://tools.ietf.org/html/rfc3986) permits only ASCII characters, thus the recommended maximum length of a SPIFFE ID is 2048 bytes.
 
-All URI components contribute to the URI length, including the "spiffe" scheme, "://" separator, trust domain name, and path component. Non-ASCII characters contribute to the URI length after they are percent encoded as ASCII characters. Note that [RFC 3986](https://tools.ietf.org/html/rfc3986) defines a maximum length of 255 characters for the "host" component of a URI; therefore a maximum length of a trust domain name is 255 bytes.
+All URI components contribute to the URI length, including the "spiffe" scheme, "://" separator, trust domain name, and path component. Note that [RFC 3986](https://tools.ietf.org/html/rfc3986) defines a maximum length of 255 characters for the "host" component of a URI; therefore a maximum length of a trust domain name is 255 bytes.
```
spikecurtis, line 116:
```suggestion
All URI components contribute to the URI length, including the "spiffe" scheme, "://" separator, trust domain name, and path component. Note that [RFC 3986](https://tools.ietf.org/html/rfc3986) defines a maximum length of 255 characters for the "host" component of a URI; therefore a maximum length of a trust domain name is 255 characters.
```

standards/SPIFFE-ID.md
```diff
@@ -109,12 +113,18 @@ Paths MAY be hierarchical - similar to filesystem paths. The specific meaning of
...
 
 ### 2.4. SPIFFE ID Parsing
 
 SPIFFE IDs follow the URI specification as defined by [RFC 3986](https://tools.ietf.org/html/rfc3986). The scheme and trust domain name of the SPIFFE ID are case-insensitive. The path is case-sensitive.
 
+### 2.5. SPIFFE ID Equivalency
+
+Two SPIFFE IDs are equivalent if and only if they match on a byte-for-byte basis. Note that since SPIFFE IDs allow only ASCII characters without percent-encoding, and SPIFFE also forbids capital letters in the `host` part of the authority (which is traditionally case-insensitive), comparisons of legal SPIFFE IDs are disambiguated.
```
spikecurtis, line 124:
```suggestion
Two SPIFFE IDs are equivalent if and only if they match on a character-for-character basis. Note that since SPIFFE IDs forbid percent-encoded characters, and forbid capital letters in the `host` part of the authority (which is traditionally case-insensitive), comparisons of legal SPIFFE IDs are disambiguated.
```

standards/SPIFFE-ID.md
```diff
@@ -185,3 +195,29 @@ For example, imagine that the operator of trust domain A and trust domain B have
...
 
+### 4.2. ID Equivalency
+The comparison of SPIFFE IDs is a security critical operation. In allowing for internationalization, as well as compatibility with the DNS system, the URI standard which SPIFFE IDs rely upon has made a handful of decisions which complicate the equivalency process. Specifically, case-insensitivity, ambiguous percent-encoding rules and support for UTF-8, and the implementation of many conditionally-special characters can make the topic a confusing one.
+
+This specification has specifically forbidden the use of characters and encoding schemes that complicate comparison (please see sections [2.1](#21-trust-domain) and [2.2](#22-path) for more information). Thus, any legal SPIFFE ID can be considered safe. It will not be misinterpreted or transformed by traditional URI libraries, and it can be safely compared with other legal IDs on a byte-for-byte basis.
```
spikecurtis, line 201:
```suggestion
This specification has specifically forbidden the use of characters and encoding schemes that complicate comparison (please see sections [2.1](#21-trust-domain) and [2.2](#22-path) for more information). Thus, any legal SPIFFE ID can be considered safe. It will not be misinterpreted or transformed by traditional URI libraries, and it can be safely compared with other legal IDs on a character-for-character basis.
```

standards/SPIFFE-ID.md
```diff
@@ -185,3 +195,29 @@ For example, imagine that the operator of trust domain A and trust domain B have
...
+1. Strip prefix "spiffe://"
+1. The (new) first character must not be a `/`
+1. From the beginning, strip one character at a time:
+    1. Ensure it is either a lowercase alpha-numeric, a `.`, a `-`, or a `/`
+    1. When `/` is detected, stop
+    1. The number of stripped characters should not exceed 255, not including `/`
+1. What remains is the path - scan it for the following conditions in sequential order:
+    1. The string does not start or end with `/`
+    1. Includes only uppercase or lowercase alpha-numeric characters, `.`s, `-`s, `_`s, and `/`s
+    1. The character sequences `//`, `/./`, and `/../` do not appear anywhere in the string
```
spikecurtis, line 220: I think this would allow something like `spiffe://mydomain/spike/..`  --- that is to say, it doesn't eliminate relative path segments as the final segment.

standards/SPIFFE-ID.md
```diff
@@ -185,3 +195,29 @@ For example, imagine that the operator of trust domain A and trust domain B have
...
+
+This specification has specifically forbidden the use of characters and encoding schemes that complicate comparison (please see sections [2.1](#21-trust-domain) and [2.2](#22-path) for more information). Thus, any legal SPIFFE ID can be considered safe. It will not be misinterpreted or transformed by traditional URI libraries, and it can be safely compared with other legal IDs on a byte-for-byte basis.
+
+It is still possible for a legal URI (but illegal SPIFFE ID) to be legally processed according to URI normalization rules _into_ a legal SPIFFE ID. Thus, under certain circumstances, it is possible for an illegal SPIFFE ID to be passed through a URI parser and produce a legal SPIFFE ID. These two SPIFFE IDs **will not match** on a byte-for-byte basis, however there exists room for confusion.
```
spikecurtis, line 203:
```suggestion
It is still possible for a legal URI (but illegal SPIFFE ID) to be legally processed according to URI normalization rules _into_ a legal SPIFFE ID. Thus, under certain circumstances, it is possible for an illegal SPIFFE ID to be passed through a URI parser and produce a legal SPIFFE ID. These two SPIFFE IDs **will not match** on a character-for-character basis, however there exists room for confusion.
```

spikecurtis (changes requested):

standards/SPIFFE-ID.md
```diff
@@ -185,3 +195,29 @@ For example, imagine that the operator of trust domain A and trust domain B have
...
+
+This is an important detail to keep in mind, particularly when accepting untrusted input that is not signed over by a trust domain authority (e.g. user input from a web app). Untrusted input should always be validated against the SPIFFE ID ruleset prior to accepting and parsing with a traditional URI parser. Please see [Appendix A](#5-appendix-a) for a brief example.
+
+## 5. Appendix A. Lightweight SPIFFE ID validation
+SPIFFE IDs have a strict character set that is designed to be as consistent and as easy to validate as possible. Here is a logic example demonstrating a simple SPIFFE ID validation mechanism:
+1. Total number of characters is less than 2048
```
spikecurtis, line 209: Since less than 2048 is only a _SHOULD_ in the standard, I don't think this can be part of the validation algorithm

standards/SPIFFE-ID.md
```diff
@@ -109,12 +113,18 @@ Paths MAY be hierarchical - similar to filesystem paths. The specific meaning of
...
 
 ### 2.4. SPIFFE ID Parsing
 
 SPIFFE IDs follow the URI specification as defined by [RFC 3986](https://tools.ietf.org/html/rfc3986). The scheme and trust domain name of the SPIFFE ID are case-insensitive. The path is case-sensitive.
 
+### 2.5. SPIFFE ID Equivalency
+
+Two SPIFFE IDs are equivalent if and only if they match on a byte-for-byte basis. Note that since SPIFFE IDs allow only ASCII characters without percent-encoding, and SPIFFE also forbids capital letters in the `host` part of the authority (which is traditionally case-insensitive), comparisons of legal SPIFFE IDs are disambiguated.
+
+It is important to try to minimize the amount of processing done on SPIFFE IDs received from untrusted sources prior to comparing them, as it is possible for some URI parsers to normalize an illegal SPIFFE ID into a legal one. Please see Security Considerations [section 4.2](#42-id-equivalency) for more information.
```
justinburke, line 126:
```suggestion
As it is possible for some URI parsers to normalize an illegal SPIFFE ID into a legal one, it is important to try to minimize the amount of processing done on SPIFFE IDs received from untrusted sources prior to comparing them. Please see Security Considerations [section 4.2](#42-id-equivalency) for more information.
```

standards/SPIFFE-ID.md
justinburke: "Thus, any legal SPIFFE ID can be considered safe. It will not be misinterpreted or transformed by traditional URI libraries, and it can be safely compared with other legal IDs on a byte-for-byte basis."

This is the intention, or aspiration of the design. But as stated, I read this as a fact. Are we actually able to make this claim? Do we know that we haven't made a mistake somewhere or that some future parser may deviate from today's norms?

standards/SPIFFE-ID.md
```diff
@@ -185,3 +195,29 @@ For example, imagine that the operator of trust domain A and trust domain B have
 
 Now imagine that a request for the customer’s data is received by the storage service in trust domain A, except the caller presented an SVID from trust domain B. Even though the presented SVID may have the necessary assertion indicating that the shared customer was authenticated and authorized the request, it is a bad idea to blindly assume that trust domain B has indeed authenticated your customer. If trust domain B is compromised, or if it has a malicious internal actor, it could claim to have authenticated any user it wishes, thus creating the very circumstance that the measure was designed to mitigate in the first place.
 
+### 4.2. ID Equivalency
+The comparison of SPIFFE IDs is a security critical operation. In allowing for internationalization, as well as compatibility with the DNS system, the URI standard which SPIFFE IDs rely upon has made a handful of decisions which complicate the equivalency process. Specifically, case-insensitivity, ambiguous percent-encoding rules and support for UTF-8, and the implementation of many conditionally-special characters can make the topic a confusing one.
```
justinburke, line 199:
```suggestion
The comparison of SPIFFE IDs is a security critical operation. In allowing for internationalization, as well as compatibility with the DNS system, the URI standard which SPIFFE IDs rely upon has made a handful of decisions which complicate the equivalency process. Specifically, case-insensitivity, ambiguous percent-encoding rules and support for UTF-8, and the implementation of many conditionally-special characters can make interpretation and implementation of the rules challenging.
```

standards/SPIFFE-ID.md
justinburke, line 203: "These two SPIFFE IDs **will not match** on a byte-for-byte basis, however there exists room for confusion."

Is it the IDs or the URLs which will not match byte-for-byte/character-for-character?

Consider rephrasing to something like: The pre- and post-processed SPIFFE IDs/URIs will not match on a ... basis.

"however, there exists room for confusion"

I don't understand this last phrase. Can it be dropped?

standards/SPIFFE-ID.md
```diff
@@ -185,3 +195,29 @@ For example, imagine that the operator of trust domain A and trust domain B have
...
+
+This specification has specifically forbidden the use of characters and encoding schemes that complicate comparison (please see sections [2.1](#21-trust-domain) and [2.2](#22-path) for more information). Thus, any legal SPIFFE ID can be considered safe. It will not be misinterpreted or transformed by traditional URI libraries, and it can be safely compared with other legal IDs on a byte-for-byte basis.
+
+It is still possible for a legal URI (but illegal SPIFFE ID) to be legally processed according to URI normalization rules _into_ a legal SPIFFE ID. Thus, under certain circumstances, it is possible for an illegal SPIFFE ID to be passed through a URI parser and produce a legal SPIFFE ID. These two SPIFFE IDs **will not match** on a byte-for-byte basis, however there exists room for confusion.
+
+This is an important detail to keep in mind, particularly when accepting untrusted input that is not signed over by a trust domain authority (e.g. user input from a web app). Untrusted input should always be validated against the SPIFFE ID ruleset prior to accepting and parsing with a traditional URI parser. Please see [Appendix A](#5-appendix-a) for a brief example.
```
justinburke, line 205: "This is an important detail to keep in mind"

What is the subject of the sentence? (What does "this" refer to?)

standards/SPIFFE-ID.md
justinburke: I could go either way on whether to include the above appendix text.

Pointing to example parsers sounds like a good idea.

standards/SPIFFE-ID.md
```diff
@@ -185,3 +195,29 @@ For example, imagine that the operator of trust domain A and trust domain B have
...
+
+## 5. Appendix A. Lightweight SPIFFE ID validation
+SPIFFE IDs have a strict character set that is designed to be as consistent and as easy to validate as possible. Here is a logic example demonstrating a simple SPIFFE ID validation mechanism:
+1. Total number of characters is less than 2048
+1. Starts with "spiffe://"
+1. Strip prefix "spiffe://"
+1. The (new) first character must not be a `/`
+1. From the beginning, strip one character at a time:
+    1. Ensure it is either a lowercase alpha-numeric, a `.`, a `-`, or a `/`
+    1. When `/` is detected, stop
```
justinburke, line 215:
```suggestion
    1. If the character is `/`, stop
```

standards/SPIFFE-ID.md
justinburke: Why is the substitution of "bytes" for "characters" important?

What are we asking implementors to do differently with characters than with bytes?

If I were to implement this spec and encountered "bytes" here, I would know to process the input string byte-by-byte. But if the spec says "characters", then that suggests to me that I need to know more about the definition of "character". Are they fixed-width? Do I need to use some particular charset?

justinburke (changes requested):

spikecurtis:
@justinburke interestingly, GitHub is letting me edit your comment, but not _reply to_ it.  So, I'll do so here

> Why is the substitution of "bytes" for "characters" important?
>
>What are we asking implementors to do differently with characters than with bytes?
>
>If I were to implement this spec and encountered "bytes" here, I would know to process the input string byte-by-byte. But if the spec says "characters", then that suggests to me that I need to know more about the definition of "character". Are they fixed-width? Do I need to use some particular charset?

A URI is defined as a sequence of _characters_, not bytes.  To quote from the RFC

> A URI is a sequence of characters from a very limited set: the letters of the basic Latin alphabet, digits, and a few special characters.  A URI may be represented in a variety of ways; e.g., ink on paper, pixels on a screen, or a sequence of character encoding octets.

So, it is simply not logically consistent to speak about the length of a URI or URI-component in terms of bytes _unless you also specify the character encoding_.  As a simple example, a URI encoded in UTF-16 will occupy twice the number of bytes as one encoded in UTF-8.

Implementers might need to think about both characters and bytes, depending on the question they are asking themselves.  Like, if I write an in-memory data-structure using SPIFFE-IDs as keys in Java, and I want to know worst case occupancy, that would be 4096-bytes to store the key, since java internally stores Strings with two bytes per character.  But, if I'm writing a SQL database schema, I can just write CHAR(2048) as the column type and not really worry too much about it.

justinburke:
> GitHub is letting me edit your comment, but not reply to it

Same here. Strange bug.

> So, it is simply not logically consistent to speak about the length of a URI or URI-component in terms of bytes unless you also specify the character encoding.

We currently have in section 2.3 of this spec (and unmodified by this PR) that we're using ASCII ("RFC 3986](https://tools.ietf.org/html/rfc3986) permits only ASCII characters"). I interpret that as specifying a character encoding. Do you disagree?

> Implementers might need to think about both characters and bytes, depending on the question they are asking themselves. Like, if I write an in-memory data-structure using SPIFFE-IDs as keys in Java, and I want to know worst case occupancy, that would be 4096-bytes to store the key, since java internally stores Strings with two bytes per character. But, if I'm writing a SQL database schema, I can just write CHAR(2048) as the column type and not really worry too much about it.

I think that what an implementation does within the confines of itself is beyond scope for this spec. What is in scope (or should be, IMO), are the requirements and expectations for interoperating between distinct implementations. What are the requirements for how an ID is represented in- and communicated via- an SVID?

The spec needs to unambiguously state what is required for the exchange and safe interpretation of SVIDs. For example, if I send you a valid, spec-conforming X.509 SVID, should you expect that the SPIFFE ID is contained in an ASN.1 Octet String, which has maximum length 2048 octets and that each octet has a one-to-one correspondence with an ASCII character?

If we change this spec to prefer 'character' instead of 'byte', then we need followup PRs to update each SVID spec so that they clearly define encoding requirements. Currently, both specs refer to this one to define the SPIFFE ID concept.

I suggest that we strongly prefer language in this spec which promotes safe implementation and strive to make it difficult for ordinary, non-security experts to incorrectly implement. So far, the distinction between 'character' and 'byte' feels to me like a distinction without  a practical difference and an unnecessary obfuscation.  If it would make the text clearer, perhaps we could include a warning to implementors that not all string libraries are the same (see Java Strings).

spikecurtis:
> We currently have in section 2.3 of this spec (and unmodified by this PR) that we're using ASCII ("RFC 3986](https://tools.ietf.org/html/rfc3986) permits only ASCII characters"). I interpret that as specifying a character encoding. Do you disagree?

I think there was a desire to express the requirements in bytes for some reason that I don't fully recall.  What we have now seems to go something like this: 

1. you must allow SPIFFE-IDs that are at least 2048 characters
2. there exists an encoding that allows you to encode URIs with one byte per character
3. therefore, you must allow at least 2048 bytes

I personally think 2 and 3 are unnecessary, and we should just stop at 1.

> The spec needs to unambiguously state what is required for the exchange and safe interpretation of SVIDs.

Here again, I think we just need to talk about characters, not bytes in this spec.  A safe SVID must support unambiguously encode a SPIFFE-ID, which itself is a string of characters.  Exactly how the SPIFFE-ID is encoded depends on the SVID type and is therefore beyond the scope of this document.

In the case of X.509, because we encode the SPIFFE-ID in a URI SAN, it is unambiguous how the SPIFFE-ID is encoded: it's an IA5String encoded in ASN.1 (https://datatracker.ietf.org/doc/html/rfc4985).

In the case of JWT, JWT is itself a text format, and so the SPIFFE-ID is encoded the same way as the rest of the document. Any ambiguity on the encoding there affects the document as a whole, so SPIFFE-ID is not specially affected.  We might need to revisit whether the spec is clear about safe exchange of these documents vis-a-vis encoding.

> I suggest that we strongly prefer language in this spec which promotes safe implementation and strive to make it difficult for ordinary, non-security experts to incorrectly implement. So far, the distinction between 'character' and 'byte' feels to me like a distinction without a practical difference and an unnecessary obfuscation.

I agree that we want to promote safe implementation for non-security experts, and I think just sticking to "characters" and not talking about "bytes" is the way forward.  String libraries in most modern languages are organized around characters, and so if we talk about about and encourage people to think about characters they should be fine. Telling people to do things like compare byte-for-byte is actually where the foot-guns lie.

azdagron (commented):
I also feel that character is the right "primitive" to talk about here, instead of bytes. As mentioned, different programming languages and libraries use different "natural" string encoding. When you are consuming things from these languages/libraries (e.g. URI SAN in an X.509 certificate), more likely than not it is going to be represented to your code using the natural string encoding. If we mandate that people do byte-by-byte comparison, then they have to do one more conversion, for really no benefit that I can see. For validation purposes, it is generally easy to ask the question "is this character ASCII" in whatever encoding is being used (e.g. does the Win32 IsTextUnicode function return IS_TEXT_UNICODE_ASCII16 to indicate that the string is zero-extended ASCII). If you are comparing two SPIFFE IDs that you have already validated, then a character by character comparison feels unambiguous and hard to get wrong.

standards/SPIFFE-ID.md
v0lkan: > Maybe we can add it to this PR and check it into a new /standards/examples directory?

Makes sense. 

I was also thinking, maybe it'd be useful to have a `validate()` function in `go-spiffe` — like a static function that takes a String and return true/false based on whether the ID is standards-conformant. It can even have a "strict" flag that treats "SHOULD"s as "MUST" (hence ensuring the length does not exceed 2048 chars—which is, normally, a recommendation, not a requirement)

PR #248, opened 2023-04-19
Add 2023H1 election dir and README

evan2645:
This PR creates the 2023 H1 election directory, and sets the README there. I basically guessed at the updated timeline because I know that the nomination period was extended but the tracking issue wasn't updated so I don't know if the dates here are accurate or not.

### DIFF
--- a/ssc/elections/2023H1/README.md
+++ b/ssc/elections/2023H1/README.md
@@ -1,0 +1,15 @@
+# 2023 H1 SSC Election
+This subdirectory captures the nominees and results of the 2023 H1 SSC election. The timeline for this election is as follows:
+* April 5th, 2023: Nominations open
+* April 19th, 2023: Nominations close
+* April 26th, 2023: Polls open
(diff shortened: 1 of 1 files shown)


amoore877 (approved):

ssc/elections/2023H1/README.md
```diff
@@ -0,0 +1,15 @@
+# 2023 H1 SSC Election
+This subdirectory captures the nominees and results of the 2023 H1 SSC election. The timeline for this election is as follows:
+* April 5th, 2023: Nominations open
+* April 19th, 2023: Nominations close
```
amoore877, line 4: I think 21st would be the close date

evan2645: Ok - is the polls open date below accurate? (26th)

amoore877: @dfeldman can you confirm if we are shifting the rest of timelines, or only the nominations?

PR #252, opened 2023-05-05
extending voting process times

amoore877:
In recent election, we ran into some issues on gathering participation and multiple other issues (ties, communication).

This PR aims to address that with a large extension of timelines for the entire process, to increase chances of participation.

This PR *does not* address additional communication gaps that were identified, ties/runoffs, or gaps in selection of eligible voters. This is intentionally not included so that we can scope discussions to just one or two topics at a time.

### DIFF
--- a/ssc/elections/README.md
+++ b/ssc/elections/README.md
@@ -12,7 +12,7 @@ All nominees are proposed via GitHub Pull Request, and each nominee gets a dedic
 
 The rest of this section captures the exact steps necessary to begin and complete an SSC election. These steps are to be performed by an SSC member, or appointed party.
 
-### Term Start-35 Days: Preparation
+### Term Start-84 Days: Preparation
 1. Identify eligible participants
 	1. Eligible participants MUST have a public email address on their GitHub profile
 1. Open GitHub issue to track election
@@ -22,7 +22,7 @@ The rest of this section captures the exact steps necessary to begin and complet
 	1. Slack #announcements channel
 	1. SIG mailing lists
 
-### T-28 Days: Nomination Opens
+### T-70 Days: Nomination Opens
 1. Create the election directory (e.g. `ssc/elections/2021H1`)
 	1. Copy in `ELECTION_README_TEMPLATE.md`, renaming to `README.md`
 		1. Fill in details, as appropriate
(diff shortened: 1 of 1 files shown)


ssc/elections/README.md
```diff
@@ -50,7 +50,17 @@ The rest of this section captures the exact steps necessary to begin and complet
 	1. Slack #announcements channel
 	1. SIG mailing lists
 
-### T-7 Days: Polling Closes and Results Announced
+### T-35 Days: Initial Polling Closes
+We add this buffer in order to account for handling ties.
+1. Check if we have a tie situation in CIVS.
+	1. If no ties, move to the next section.
+	1. For non-tied seats, move to the next section *just for that seat*.
+	1. For tied seats:
```
evan2645: I think there are some optics around knowing who tied and who didn't, because at some level you can deduce which place people fell into in the results. In the past, we put some effort into trying to obscure exact vote numbers/rankings etc because it can induce some weird social situations, and make people uncomfortable.

Overall, my two cents here is that we should do what we can to avoid signaling in this way .. it may be difficult to avoid (and perhaps not worth the potential cost), but thought it was worth calling out as a consideration. One possibility could be to not announce winner in a partial tie until the end, another one could be to change the CIVS voting algorithm to better protect result privacy in the face of a tie, etc.

evan2645 (commented):
thank you @amoore877 , this is much better than what we had before. I left a small thought that is good to keep in mind but doesn't need to be addressed right now.

ssc/elections/README.md
amoore877: I am unsure how to handle ties in a multi-seat election but where at least one seat clearly was one. It feels unfair to me to have the clear winner go back into the runoff, and additionally I think (assuming people's preferences are unchanged) would be likely to result in the same situation again just magnified- the one person clearly winning and the others tying. Condorcet voting actually somewhat guarantees that to be the case by design.

A potential change to just avoid the problem completely would be to have 0 elections where more than one seat is up for voting. Then there is no such thing as "X clear seat winners, and Y tied seats to do runoff".
It was also brought up in discussion that we could instead say all such people win a seat, with a future de-scaling of the seats later. Though my fear there is a tie, especially a conspiratorial one, resulting in a doubling+ of seats.

v0lkan: related thought: raising awareness and having more people to vote will reduce (but not eliminate) the possibility of ties.

--

I am even thinking whether it makes sense to assign ties as co-chairs. 

I think 3+ people ties will be "very" exceptional; and we may have a special clause for that.

And if there is a 2-people tie, they can carry the responsibility together.

Just thinking out loud.

v0lkan: I mean if a "purely mathematical" solution has social (as per Evan's comment: "optical") consequences, we can come up with a still optimal solution by relaxing certain requirements.

--
(as a sidenote: that's also how most of those leetcode-hard problems are generated:
pick an NP-hard problem; update constraints so that it is not np-hard given certain conditions; ask it to the candidate and watch the candidate act as if they have never seen that problem before -- but I digress )

amoore877: so I think we need to discuss within SSC something that will heavily simplify outcomes- splay out our terms so there is never more than one seat up for election at once.

that will remove one whole logic branch in this process, and if ties do happen it removes the "hard feelings" concern that has been brought up where there's a clear winner announced but a runoff for those that may feel like "losers that tied" in that scenario.

we can also decide then on just the one logic branch dimension- what to do in event of ties for that one and only seat up for grabs.

speaking of simplification, I see no comments yet around non-tie changes I have made.

to drive iterative improvement here and keep focus in following PRs, I'll alter this PR to just have the timeline updates. we can discuss in SSC sync about:
- ties
- comms
- whatever else
and I'll do a PR for each as we reach agreements (or close enough agreements for a PR to exist)

v0lkan (approved):
LGTM!

Issue #258, opened 2023-08-16
2023H2 SSC Election


dfeldman:
### 2023 H2 SSC Election
This is the official tracking issue for the 2023H2 SSC election. The timeline for this election is as follows:
* AUGUST 23, 2023: Nominations open
* SEPTEMBER 6, 2023: Nominations close
* SEPTEMBER 13, 2023: Polls open
* OCTOBER 11, 2023: Polls close
* NOVEMBER 1, 2023: Results announced

### How to Participate
All SPIFFE community members and contributors demonstrating active engagement in the project(s) are invited to both nominate and vote on new SSC members. An initial batch of eligible participants discovered via automation is listed below. **However, if you are not on the list below, but you participate in the project, you are still eligible to be nominated and to vote.**

For more information about the definition of active engagement, and how this list was compiled, please see the Election and Term Mechanics section of the [SSC Charter](https://github.com/spiffe/spiffe/blob/master/ssc/CHARTER.md#election-and-term-mechanics).

If you feel that you are an active SPIFFE community member or contributor but are not included in the list below, please contact the SSC at [ssc-elections@spiffe.io](mailto:ssc-elections@spiffe.io) and cc [dfeldman.mn@gmail.com](mailto:dfeldman.mn@gmail.com) and we will be happy to include you. Even if you do not interact with GitHub normally, to be an election candidate a GitHub account is required as this is how SSC members gain administrative rights on the SPIFFE projects.

### Nominating an SSC Member
Eligible participants may nominate up to two candidates per available SSC seat during the nomination period. They may nominate themselves or someone else.

If you'd like to nominate an SSC member for this election cycle, please follow these steps:

Verify that your GitHub handle is included in the list of eligible participants below
Verify that, in your best judgement, the nominee meets the criteria specified in the Nominee Qualifications section of the [SSC Charter](https://github.com/spiffe/spiffe/blob/master/ssc/CHARTER.md#nominee-qualification)
Fork this repository
Copy ssc/elections/NOMINEE_TEMPLATE.md into the appropriate subdirectory, and name it after the person to be nominated
For example, from the root directory:
$ cp ssc/elections/NOMINEE_TEMPLATE.md ssc/elections/2021H1/JANE_DOE.md
Fill in all fields in the copied template as completely as possible
Open a GitHub Pull Request back to this repository to add the new nominee
Create a new commit with the name of the election and nominee
For example, Nominate Jane Doe for 2023H2 SSC Election
[...]

PR #267, opened 2023-12-11
fix Mohamed Omar's election entry for GitHub

amoore877:
Typo was made during election issue #258 , putting wrong GH entry for candidate (who is now elected)


v0lkan (approved):
🌯

PR #273, opened 2024-03-12
add more bot account skips to voter discovery

amoore877:
We've introduced more bots that should not be allowed to vote.


v0lkan (approved):
🌮 .

PR #274, opened 2024-03-12
consolidate email/eligibility instructions in election template

amoore877:
Rather than have two potentially long lists in this already large issue template, consolidate to just marking the usernames that need attention.


v0lkan (approved):
🌮 .

PR #275, opened 2024-03-12
Election nominations immediate

amoore877:
Rather than lag between initial announcements and open nominations, just let nominations occur while announcements are fresh.


v0lkan (approved):
🌮 .

PR #287, opened 2024-05-24
SSC members as slack owners, not just admins

amoore877:
We should later develop who, if anyone, should be Admins in SPIFFE slack.

For now, it makes sense that SSC would be Owners.

Related to recent effort on removing Owner from inactive/departing non-SSC.


v0lkan (approved):
🌮

PR #291, opened 2024-07-08
add note about donating tiny projects

amoore877:
Set an initial guidance for how we handle very tiny projects (plugins, small libraries, maybe groups of very helpful scripts...), as opposed to full-on services or systems.


kfox1111 (commented):
LGTM. Thanks! :)

amoore877:
Will leave this open for a day in case other SSC members have comments

kfox1111 (approved):

PR #296, opened 2024-07-31
Fix broken links and minor typo

heymarcel:
Fixed a few more broken links while reading the spec. Also fixed a minor typo (compound modifier missing hyphen)


azdagron (approved):
Thanks @heymarcel !

PR #304, opened 2024-09-05
Introduce SPIFFE Bundle Map spec

evan2645:
This PR introduces the SPIFFE Bundle Map text that has been making its
way through SIG-Spec. A link to that document is below for posterity.

https://docs.google.com/document/d/13KHycYIfRC-g42aEfo4_4inF_WauCXOBgZAcUGuxdgs

### DIFF
--- a/standards/SPIFFE_Trust_Domain_and_Bundle.md
+++ b/standards/SPIFFE_Trust_Domain_and_Bundle.md
@@ -227,7 +227,7 @@ The following is a sample SPIFFE Bundle Map:
 {
   “trust_domains”: {
     "example.com": {
-      “sequence_number”: 12035488,
+      “spiffe_sequence”: 12035488,
       "keys": [
         {
           "kty": "RSA",
@@ -250,7 +250,7 @@ The following is a sample SPIFFE Bundle Map:
 ```
 
 ### Appendix B.1. Server/Client Behavior
-SPIFFE Bundle Maps are designed to be loaded and ingested atomically. The entirety of its contents should be loaded by the consumer, and should be considered “state of the world” at this time. If measurement of propagation delay, confirmation of liveness etc, is required, consumers should use the “sequence_number” of the relevant bundles as many failure modes can cause liveness divergence across the bundles represented in the map.
+SPIFFE Bundle Maps are designed to be loaded and ingested atomically. The entirety of its contents should be loaded by the consumer, and should be considered “state of the world” at this time. If measurement of propagation delay, confirmation of liveness etc, is required, consumers should use the “spiffe_sequence” of the relevant bundles as many failure modes can cause liveness divergence across the bundles represented in the map.
 
 [1]: https://github.com/spiffe/spiffe/blob/master/standards/SPIFFE-ID.md#2-spiffe-identity
 [2]: https://github.com/spiffe/spiffe/blob/master/standards/SPIFFE-ID.md#3-spiffe-verifiable-identity-document
--- a/standards/SPIFFE_Trust_Domain_and_Bundle.md
+++ b/standards/SPIFFE_Trust_Domain_and_Bundle.md
@@ -44,7 +44,7 @@ A SPIFFE bundle is an object containing a trust domain's cryptographic keys. The
 
 SPIFFE bundles are designed for use within and between SPIFFE control plane implementations. They are not meant for direct consumption by workloads, however this specification does not preclude such use.
 
-When storing or otherwise managing SPIFFE bundles, it is important to independently record the name of the trust domain that the bundle represents, nominally through the use of a `<trust_domain_name, bundle>` tuple. When validating an SVID, validators must choose the bundle corresponding to the trust domain that the SVID resides in, so maintaining this relationship is required in most scenarios.
+When storing or otherwise managing SPIFFE bundles, it is important to independently record the name of the trust domain that the bundle represents, nominally through the use of a [SPIFFE Bundle Map](#5-spiffe-bundle-map) or via a `<trust_domain_name, bundle>` tuple. When validating an SVID, validators must choose the bundle corresponding to the trust domain that the SVID resides in, so maintaining this relationship is required.
 
 Note that the contents of a trust domain's bundle are expected to change over time as the keys it contains are rotated. Keys are added and revoked by issuing a new bundle with new keys included and revoked keys omitted. It is the responsibility of the SPIFFE implementation to distribute bundle content updates to workloads as needed. The exact format and method by which these updates are delivered is out of scope for this specification. Please see the [SPIFFE Workload API][3] specification for information on delivering updates to workloads.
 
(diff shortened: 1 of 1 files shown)


kfox1111:
LGTM! :)

standards/SPIFFE_Trust_Domain_and_Bundle.md
```diff
@@ -197,6 +220,38 @@ Trust bundle #2 for example.com:
...
 In trust bundle #2, note that the `spiffe_sequence` parameter has been incremented and the second root certificate for `example.com` has been added. Once this new trust bundle is published and distributed, validators will accept SVIDs signed by either the original or the replacement root certificate. By publishing the replacement certificate well ahead of the expiration of the original certificate, validators have ample opportunity to refresh the trust bundle of example.com and learn of the pending replacement certificate.
 
+## Appendix B. SPIFFE Bundle Map Example
+The following is a sample SPIFFE Bundle Map:
+
+```
+{
+  “trust_domains”: {
+    "example.com": {
+      “sequence_number”: 12035488,
```
yt2985: Should this be the `sequence_number` or the `spiffe_sequence`?

standards/SPIFFE_Trust_Domain_and_Bundle.md
```diff
@@ -78,15 +81,32 @@ The `kty` parameter MUST be set, and its behavior follows [Section 4.1][6] of RF
...
+## 5. SPIFFE Bundle Map
+A SPIFFE Bundle Map is an object containing a collection of SPIFFE Bundles. The bundles it contains are considered authoritative for the corresponding trust domain name, which is also included.
+
+SPIFFE Bundle Maps are designed for use between SPIFFE control plane implementations, between SPIFFE implementations and the workloads they serve, as well as internally (i.e. as an implementation detail of a SPIFFE control plane).
+
+### 5.1. SPIFFE Bundle Map Format
+SPIFFE Bundle Maps are JSON-encoded structures, encapsulating one or more SPIFFE Bundles and some additional metadata. For an example of a complete SPIFFE Bundle Map, please see [Appendix B. SPIFFE Bundle Map Example](#appendix-b-spiffe-bundle-map-example).
+
+#### 5.1.1. Trust Domains
+A static “trust_domains” key MUST be set, and MAY be empty. Its contents are a map of SPIFFE Bundles, keyed by the trust domain name. For more information on what constitutes a valid trust domain name, please see [Section 2][1] of the SPIFFE ID specification.
```
arndt-s, line 93: I'm thinking if something like "produced_at" with a timestamp value would be very helpful. If delivered asynchronously it'll be very hard to get a notion of freshness of it. A timestamp would go a long way for debugging purposes.

Alternatively, explicitly allowing additional keys to be present could make something like it an implementers choice.

standards/SPIFFE_Trust_Domain_and_Bundle.md
```diff
@@ -95,6 +115,9 @@ Continuing the above example where a naïve implementation imports a particular
...
 
+### 6.3. JSON Key Duplication
+[RFC 8259][12], which serves as the standard for JSON data format, states that the names within an object SHOULD be unique. Furthermore, it emphasizes that when the names within a JSON object are not unique, the behavior of software that receives such an object is unpredictable. This behavior poses an unacceptable security risk when the JSON format is used to represent SPIFFE Bundle Maps, as the presence of duplicate trust domain names could lead to the incorrect bundle being selected for use in validating SVIDs. As such, it is important that libraries parsing SPIFFE bundle maps employ duplicate key detection.
+
```
PieterKas, line 120: SPIFFE bundles are JWK sets and RFC7515 which gives guidance on key provenance and trust ( https://datatracker.ietf.org/doc/html/rfc7517#section-9.1). 

Should there be similar considerations for SPIFFE bundle maps to regarding their origin and integrity?

For example, are there any risks that may result from an adversary adding or removing SPIFFE bundles to a Bundle Map? Are there precautions the recipient should take?

cesarghali:
Hi @evan2645, I'm curious of there are any updates about the status of this PR. Also, when do you think it can be merged? Thank you!

evan2645:
Hi @cesarghali!

We have a SIG-Spec call scheduled tomorrow related to this PR and some other work. Anybody who intends to use this spec should join as the primary hesitance thus far is around lack of community feedback. Please see [SIG-Spec](https://github.com/spiffe/spiffe/blob/main/community/sig-spec/README.md) calendar for call information. More than happy to hold follow-on calls as well if the time doesn't work well for you.

cesarghali:
Thank you @evan2645 for the information. I will attend the SIG meeting tomorrow!

standards/SPIFFE_Trust_Domain_and_Bundle.md
yamandabbagh: Thank you @arndt-s, I agree having a timestamp has some value. We circulated this idea of having a timestamp or some sort of a refresh hint to the map during the Spiffe community meetings. The conclusion we reached that while adding something to show freshness has some value, keeping the Spec simple and leaving this detail up to the implementers would keep the specification simple and hence increase adoption. 

If we add to the spec a required timestamp/freshness field we would need an agreement about how this field would be consumed, when maps expire, how often we need to refresh. And just adding an optional field without further details and why it is important would add complexity to the Spec with minimal benefit. What do you think?

standards/SPIFFE_Trust_Domain_and_Bundle.md
yamandabbagh: That's an important point. The same trust considerations that apply to a single SPIFFE bundle, where an adversary could manipulate the contents, should also apply to the bundle map.  Both producers and consumers of these bundles must implement measures to ensure the integrity of the bundle map. The specifics of these measures will vary depending on the implementation.  To address this, we could add a security consideration to the specification emphasizing the importance of bundle map integrity, while leaving the implementation details to developers. What are your thoughts on this approach?

cesarghali:
Overall LGTM! Thanks everyone for working on this PR. Merging it would unblock planned implementations and avoid short-term and long-term divergence from the specs.

standards/SPIFFE_Trust_Domain_and_Bundle.md
arndt-s: Agreed on the focus of keeping it simple to easy use and adoption.

Closing this comment. 
Update: looks like I can't close it. Please consider closed.

standards/SPIFFE_Trust_Domain_and_Bundle.md
evan2645: It should be `spiffe_sequence`

standards/SPIFFE_Trust_Domain_and_Bundle.md
evan2645: Thank you for the observation @PieterKas .. I feel it's an important one. We've discussed this in the past when designing SPIFFE Bundle spec , but I think now things have changed with the introduction of this spec and other patterns. We probably need to have a closer look at the bundle spec as well. Could you please open a dedicated issue so we can discuss further and more broadly?

PieterKas: Sounds good to me.

evan2645:
Thanks to all who contributed to the development of this spec. I think it is ready to go. @azdagron please take a look

azdagron (approved):
The proposed addition looks good!

PR #316, opened 2025-01-21
Editorial changes to X509-SVID

arndt-s:
Small editorial changes to X509-SVID spec.
* Changed `MAY` to `SHOULD` that a signing SVID should reside in the same trust domain. The `MAY` conflicts with a `SHOULD` in SVID-Trust.
* The scheme part in the URI would be only the part before the colon.

Credits go to Mike from spiffe-slack for finding these!
- https://spiffe.slack.com/archives/C7AMR9NKB/p1737349938049169
- https://spiffe.slack.com/archives/C7AMR9NKB/p1737431930473339


performantdata (approved):

evan2645:
Thank you @performantdata and @arndt-s!

@performantdata raised [another error](https://spiffe.slack.com/archives/C7AMR9NKB/p1737172920144709) in the Slack, a stale reference to percent-encoding in section 2.3. Can we fix that up with this PR too?

evan2645 (approved):

Issue #315, opened 2025-01-19
Support for WIMSE WITs


yaronf:
The [IETF WIMSE working group](https://datatracker.ietf.org/group/wimse/about/) defines a set of protocols for workload authentication/authorization. One of these protocols is the [service-to-service protocol](https://datatracker.ietf.org/doc/draft-ietf-wimse-s2s-protocol/) that supports three methods:
1. Mutual TLS
2. A proof-of-possession method, similar to OAuth DPoP
3. And HTTP Message Signatures.

Methods (2) and(3) both require a token called WIT (Workload Identity Token) which is a slight extension of the SPIFFE JWT-SVID. The core difference is that the WIT contains a public key whose private key is held by the workload. This enables to cryptographically bind service calls to their sender.

Without going into implementation details, the S2S team is requesting that the SPIFFE API be extended to support provisioning of WITs to workloads.

A few more points:
- The request should be extended with a public key, so that the secret key never leaves the workload.
- We need to discuss cryptographic agility, i.e. being able to use different cryptographic algorithms within the token.
- WIMSE S2S is still a draft and will be changing for a while. We shouldn't wait for the RFC to be published, but then again, we may want to tag requests/responses with a draft number to prevent protocol ossification.

strideynet:
> The request should be extended with a public key, so that the secret key never leaves the workload.

This is fairly interesting to me since as it stands today, this is not possible for X509 SVID issuance via the SPIFFE Workload API. I don't think this is necessarily a bad idea, but, for the sake of consistency, we'd probably want to consider whether we should extend the X509 SVID profile to similarly allow the submission of a key-pair controlled by the workload. It would seem odd to me if this ability was available for WIT SVIDs and not X509 SVIDs.

I think this then raises another question - should we make this submission of this public key for a WIT optional or required? 

---

A bit of a tangent - but - in at least a few conversations with customers, we've heard the idea of workloads owning their own key-pair come up a few times - especially in regards to environments where some kind of HSM/PKCS11 style device is usually leveraged for key-pair material. So perhaps this is a good time to discuss the topic as a whole for SPIFFE.

arndt-s:
I'm seeing a few impacts if we would make these new key submission parameters required:
- It would be a breaking change for SPIFFE and the SPIFFE-enabled workloads of today wouldn't work with the SPIFFE spec of tomorrow.
[...]

PR #327, opened 2025-02-21
WIT-SVID Workload API Profile

strideynet:
Introduces the WIT-SVID Profile for the SPIFFE Workload API.

Part of https://github.com/spiffe/spiffe/issues/315


standards/SPIFFE_Workload_API.md
```diff
@@ -360,6 +366,122 @@ When interfacing with legacy JWT validators, the `FetchJWTBundles` method can be
 
 The `FetchJWTBundles` method returns bundles keyed by the SPIFFE ID of the trust domain. When validating a JWT-SVID, the validator MUST use the bundle corresponding to the trust domain of the subject. If a JWT bundle for the specified trust domain is not present, then the token is untrusted.
 
+## 7. WIT-SVID Profile
+
+The WIT-SVID Profile of the SPIFFE Workload API provides a set of gRPC methods which can be used by workloads to retrieve [WIT-SVIDs](WIT-SVID.md) and their related trust bundles. This profile outlines the signature of these methods, as well as related client and server behavior.
+
+Unlike the X.509-SVID and JWT-SVID profiles, implementation of the WIT-SVID profile is optional. Where a server does not support the WIT-SVID profile, calls to all RPCs defined by the WIT-SVID profile MUST return the "Unimplemented" gRPC status code. If the client receives an "Unimplemented" status code calling a WIT-SVID profile RPC, it SHOULD NOT attempt to reconnect to the server using the WIT-SVID profile.
```
strideynet, line 368: > If the client receives an "Unimplemented" status code calling a WIT-SVID profile RPC, it SHOULD NOT attempt to reconnect to the server using the WIT-SVID profile.

Does prohibiting any form of backoff seem too restrictive here?

standards/SPIFFE_Workload_API.md
```diff
@@ -360,6 +366,122 @@ When interfacing with legacy JWT validators, the `FetchJWTBundles` method can be
...
+
+### 7.1 Profile Definition
+
+The WIT-SVID Profile RPCs and associated messages are defined below. For the complete Workload API service definition, see [workloadapi.proto](workloadapi.proto).
+
+```protobuf
+service SpiffeWorkloadAPI {
```
strideynet, line 375: I took a re-read of the SPIFFE Workload Endpoint specification - and within that - we do recommend that extensions to the endpoint be achieved by adding new gRPC services rather than modifying existing ones. One of the benefits claimed is that it makes it clear via gRPC reflection whether or not functionality is supported. This has me wondering if we should consider implementing the WIT-SVID profile as its own gRPC service. WDYT?

standards/SPIFFE_Workload_API.md
```diff
@@ -360,6 +366,122 @@ When interfacing with legacy JWT validators, the `FetchJWTBundles` method can be
...
+
+The `WITBundlesRequest` request message is currently empty and is a placeholder for future expansion.
+
+The `WITBundlesResponse` response message consists of a mandatory `bundles` field, which MUST contain at least the trust bundle for the trust domain in which the server resides.
+
+If the client is not entitled to receive any WIT bundles, then the server SHOULD respond with the "PermissionDenied" gRPC status code (see the [Error Codes](SPIFFE_Workload_Endpoint.md#6-error-codes) section in the SPIFFE Workload Endpoint specification for more information). The client MAY attempt to reconnect with another call to the `FetchWITBundles` RPC after a backoff.
+
+As mentioned in [Stream Responses](#43-stream-responses), each `WITBundleResponse` response contains the complete set of authorized trust bundles for the client at that point in time. As such, if the server redacts bundles from a subsequent response (or all bundles, i.e., returns a "PermissionDenied" gRPC status code) the client SHOULD cease using the redacted bundles.
+
```
strideynet, line 483: The X509-SVID Profile includes a section titled "Default Identity" which it seems could apply almost verbatim to the WIT-SVID Profile. I'm split between:

1. Duplicating this section for the WIT-SVID profile
2. Extracting this section from the X509-SVID profile, and referencing it within the WIT-SVID and X509-SVID profile

Keen to hear any opinions on this

standards/SPIFFE_Workload_API.md
yaronf: IMO it is the right behavior for Unimplemented.

standards/SPIFFE_Workload_API.md
yaronf: Note that for JWT-SVID this section is missing, even though the `hint` field is included. This seems to point towards your second option.

standards/SPIFFE_Workload_API.md
strideynet: Yeah - that does seem to be my gut feeling.

strideynet:
I've changed the base here to merge into the `draft-wit-svid` branch - as discussed in SIG-SPEC, this branch will act as a holding point for the WIT-SVID changes to the SPIFFE specification until there is consensus that the WIMSE W2W draft is stable. We'll also be able to use `draft-wit-svid` to perform a final review before "publishing" the WIT-SVID specs.

standards/SPIFFE_Workload_API.md
```diff
@@ -27,18 +27,24 @@ Portable and interoperable cryptographic identity for networked workloads is per
...
+Currently, there are three profiles:
 
 - [X.509-SVID Profile](#5-x509-svid-profile)
 - [JWT-SVID Profile](#6-jwt-svid-profile)
+- [WIT-SVID Profile](#7-wit-svid-profile)
 
-Both profiles are mandatory and MUST be supported by SPIFFE implementations. However, operators MAY administratively disable a specific profile in their deployment.
+The X.509-SVID and JWT-SVID profiles are mandatory and MUST be supported by SPIFFE implementations. However, operators MAY administratively disable a specific profile in their deployment.
+
+The WIT-SVID Profile is optional and MAY be supported by SPIFFE implementations.
```
sorindumitru, line 52: Looking at https://spiffe.io/docs/latest/spiffe-about/overview/ it seems that not all implementation have supported both profiles. I think we should just specify that if a profile is not supported, the RPCs for it should return unimplemented.

standards/SPIFFE_Workload_API.md
sorindumitru: I think that's mostly about adding new functionality that is not port of the WorkloadAPI, so I think we should keep this profile as part of the workloadapi service.

standards/SPIFFE_Workload_API.md
```diff
@@ -360,6 +366,122 @@ When interfacing with legacy JWT validators, the `FetchJWTBundles` method can be
...
+
+The `WITBundlesRequest` request message is currently empty and is a placeholder for future expansion.
+
+The `WITBundlesResponse` response message consists of a mandatory `bundles` field, which MUST contain at least the trust bundle for the trust domain in which the server resides.
+
+If the client is not entitled to receive any WIT bundles, then the server SHOULD respond with the "PermissionDenied" gRPC status code (see the [Error Codes](SPIFFE_Workload_Endpoint.md#6-error-codes) section in the SPIFFE Workload Endpoint specification for more information). The client MAY attempt to reconnect with another call to the `FetchWITBundles` RPC after a backoff.
+
+As mentioned in [Stream Responses](#43-stream-responses), each `WITBundleResponse` response contains the complete set of authorized trust bundles for the client at that point in time. As such, if the server redacts bundles from a subsequent response (or all bundles, i.e., returns a "PermissionDenied" gRPC status code) the client SHOULD cease using the redacted bundles.
```
sorindumitru, line 482: Should we support some kind of pagination API here (and in the other profiles)? gRPC has some default receive message sizes (I think it's 4MB). If you have large bundles or large number of federation relationships, you may end up not being able to receive the updates anymore.

standards/SPIFFE_Workload_API.md
strideynet: Yes - I think that seems like a reasonable change to make to better reflect reality here.

standards/SPIFFE_Workload_API.md
strideynet: Hm - it's a good point. We should discuss this in next SIG-Spec since I think I think modifying the existing profiles will merit further discussion esp in regards to back/fore compatibility.

standards/SPIFFE_Workload_API.md
```diff
@@ -360,6 +366,122 @@ When interfacing with legacy JWT validators, the `FetchJWTBundles` method can be
...
+The WIT-SVID Profile RPCs and associated messages are defined below. For the complete Workload API service definition, see [workloadapi.proto](workloadapi.proto).
+
+```protobuf
+service SpiffeWorkloadAPI {
+  /////////////////////////////////////////////////////////////////////////
+  // WIT-SVID Profile
+  /////////////////////////////////////////////////////////////////////////
+
+  // Fetch WIT-SVIDs for all SPIFFE identities the workload is entitled to,
+  // as well as related information like trust bundles. As this information
```
evan2645: > as well as related information like trust bundles

There is a similar comment on the `WITSVID` message, but I don't see where a bundle is delivered as part of this response. I don't think we need one anyways due to the asymmetric nature of WIT auth ... I checked the JWT-SVID profile and don't see one there either.

standards/SPIFFE_Workload_API.md
arndt-s: agreeing with Soren here. Also, gRPC reflection will still tell one if a certain RPC is available, that won't help for situations though where the full Workload API is present but RPCs return unimplemented.

PS: Just checked and it seems that a server also returns UNIMPLEMENTED when it doesn't know about the RPC (https://grpc.io/docs/guides/error/#general-errors) which gets me thinking of we can make us of that and say that WIT RPCs should only be part of the protobuf if supported? Aka discoverable by reflection?

standards/SPIFFE_Workload_API.md
```diff
@@ -360,6 +366,122 @@ When interfacing with legacy JWT validators, the `FetchJWTBundles` method can be
...
+  // WIT-SVID profile.
+  rpc FetchWITBundles(WITBundlesRequest) returns (stream WITBundlesResponse);
+
+  // ... RPCS for other profiles ...
+}
+
+// The WITSVIDRequest message conveys parameters for requesting a WIT-SVID.
+// There are currently no such parameters.
+message WITSVIDRequest {
+  // TODO(noah): Eventually, we may consider introducing workload key generation
```
arndt-s: For the sake of merging this into a "preview" branch I would remove that.

arndt-s (approved):

standards/SPIFFE_Workload_API.md
strideynet: 👍  Yeah - I think in my original draft I had this returning the trust bundle but after discussion we removed that - looks like I forgot to update this comment.

standards/SPIFFE_Workload_API.md
strideynet: Addressed in 512964d

standards/SPIFFE_Workload_API.md
strideynet: Addressed in 62c51f3

standards/SPIFFE_Workload_API.md
strideynet: > Aka discoverable by reflection?

This is an interesting idea to me and I can see the merit in it, but, I do think this could be awkward in some ways. Today, an implementation of a SPIFFE Workload API in Go can directly use the generated Go Protobuf bindings from the `go-spiffe` repository. However, if we are to say that you should only use the protobuf containing these if you implement the WIT profile, then they must generate their own Go Protobuf bindings (unless we were to publish a generated binding with/without the RPCs). This process would only become more complex in the future if we were to introduce additional profiles.

standards/SPIFFE_Workload_API.md
strideynet: As discussed in SPIFFE SPEC, we'll hold off on changing the wording for the JWT/X509 profiles for now and just state that the WIT profile is optional.

standards/SPIFFE_Workload_API.md
strideynet: We roughly found consensus at SIG-Spec that we should introduce pagination for the new profile - but we did not reach consensus on what that will precisely look like.

I will reach out to @arndt-s since we'll want to make sure pagination works the same across both of our new APIs.

strideynet: I have included a proposed design for pagination in [713f75e](https://github.com/spiffe/spiffe/pull/327/commits/713f75e3465d545b99003fbd7c24f71d57f64685).

strideynet:
I have included a proposed design for pagination within 713f75e - I would appreciate a review of this change. Myself and @arndt-s have discussed this, and it is likely a similar mechanism will be introduced in the Broker API specification.

I do have an underlying feeling that the pagination mechanism introduces a significant increase in the complexity of the RPC - but perhaps this is just in contrast to the simplicity of the other RPCs where the client can expect each message to be the full set.

standards/SPIFFE_Workload_API.md
strideynet: Addressed in eeaa9ba

standards/SPIFFE_Workload_API.md
```diff
@@ -495,6 +485,16 @@ If the client is not entitled to receive any WIT bundles, then the server SHOULD
 
 As mentioned in [Stream Responses](#43-stream-responses), each complete set of response messages within an update contains the complete set of authorized trust bundles for the client at that point in time. As such, if the server redacts bundles from a subsequent update (or all bundles, i.e., returns a "PermissionDenied" gRPC status code) the client SHOULD cease using the redacted bundles.
 
+### 8. Default Identity
```
yaronf: I think you meant a double-#.

yaronf:
> I have included a proposed design for pagination within [713f75e](https://github.com/spiffe/spiffe/commit/713f75e3465d545b99003fbd7c24f71d57f64685) - I would appreciate a review of this change. Myself and @arndt-s have discussed this, and it is likely a similar mechanism will be introduced in the Broker API specification.
> 
> I do have an underlying feeling that the pagination mechanism introduces a significant increase in the complexity of the RPC - but perhaps this is just in contrast to the simplicity of the other RPCs where the client can expect each message to be the full set.

I'm probably missing some context, please enlighten me: do we expect more trust bundles in the WIMSE space compared to regular SPIFFE trust bundles? Otherwise, why define pagination specifically for WIT trust bundles? Wouldn't it make more sense to play with this idea in the more mature JWT-SVID space - where we're already seeing large scale deployments - and only later apply it to WITs?

strideynet:
Following discussion in SIG-Spec, the general consensus seems to be that we ought to kick the issue of pagination down the road. It introduces significant complexity to the spec and client/server implementations and feels like a topic worth its own deeper discussion. Given that we would need to handle back/fore compatibility for the JWT/X509 profiles, it doesn't seem like there's a significant advantage to trying to introduce this now for the WIT profile.

As such, I have reverted the commit introducing pagination.

standards/SPIFFE_Workload_API.md
strideynet: 86d1778

standards/WIT-SVID.md
```diff
@@ -0,0 +1 @@
+Placeholder
```
sorindumitru: Is this waiting for the IETF side of things to conclude before committing to something? We might want to at least point towards the current draft for details about the WIT-SVID.

sorindumitru (commented):
We also discussed having this as an experimental feature first. It would be good to point that out in the documents.

standards/WIT-SVID.md
strideynet: Ah - this was a placeholder since I think we should write this page in a separate PR - but I wanted to be able to link to it from this draft.

strideynet:
> We also discussed having this as an experimental feature first. It would be good to point that out in the documents.

Yeah - do you think perhaps a header on the spec stating that this is an "experimental" branch spec is the right way of doing this ?

sorindumitru:
> Yeah - do you think perhaps a header on the spec stating that this is an "experimental" branch spec is the right way of doing this ?

Would that be in WIT-SVID.md? That sounds ok to me. If you do that, I think it would be useful to also link to the latest draft or the repo with the draft. Because otherwise it's hard to tell what exactly the WIT-SVID is supposed to be.

standards/SPIFFE_Workload_API.md
```diff
@@ -360,6 +356,129 @@ When interfacing with legacy JWT validators, the `FetchJWTBundles` method can be
...
+}
+```
+
+### 7.2 Profile RPCs
+
+#### 7.2.1 FetchWITSVID
+
+The `FetchWITSVID` RPC streams back WIT-SVIDs. 
+
+The `WITSVIDRequest` request message is currently empty and is a placeholder for future expansion.
```
arndt-s: I know we discussed this before but I don’t remember, what was the reason we don’t support requesting a specific SPIFFE-ID here?

arndt-s (approved):
One comment that is probably more a question, apart from that I believe we’re good to go

standards/SPIFFE_Workload_API.md
strideynet: If I recall correctly, the main reason for it's inclusion on the JWT-SVID profile is because the act of requesting a JWT-SVID will likely cause a cryptographic operation to occur since the JWT-SVIDs may have varying audiences and therefore, they are not easy to cache. Hence, the ability to request a specific SPIFFE ID avoids signing N JWT SVIDs when the client will likely only use 1. Since this limitation is not present for WIT-SVIDs, then it is more similar in this regard to X509-SVIDs for which the profile does not support this.

I have a few thoughts beyond that:

- I don't see it as overly harmful to include if we wanted, but, I do think it does somewhat weaken the purpose of the `hint` field. In how many cases should we expect a workload to know apriori what its SPIFFE ID will be?
- If we don't add it now, it will definitely be harder to add at a later date since a client cannot know if the server will respect the request ahead of time. Client implementations would need to manually account for a server that does not support the field.

yaronf: @strideynet Isn't the whole point of a `hint` field that the sender cannot assume the receiver understands/respects it? Or are you saying that a server that does not recognize this field will return an error?

strideynet: > Isn't the whole point of a hint field that the sender cannot assume the receiver understands/respects it?

Yeah - that's correct. I suppose my point here is that today for X509-SVIDs, the "hint" is  the "recommended" mechanism for a client to select which SVID it should use when entitled to multiple. If we add a mechanism whereby the client can ask the server for a specific SPIFFE ID, then that's a second mechanism which could be used for a similar purpose. I'm not sure in reality it makes a significant difference, but, it does seem to mark a departure from the idea that a workload will not typically know its SPIFFE ID prior to requesting one.

Equally, I do think there are advantages to adding it - in the SPIRE implementation, the agent which serves the Workload API tends to have the cache pre-filled with SVIDs that a workload may receive - however, there are other implementations of SPIFFE where this is not the case, and in these cases, if a workload can directly request a specific SPIFFE ID then this could save the issuer performing unnecessary work to issue SVIDs that the workload does not intend to use.

arndt-s:
@SPIFFE maintainers (@evan2645 , @justinburke, @spikecurtis , @azdagron ) - we believe this PR is ready to be merged into the **preview** branch. Can someone of you please review?

strideynet:
I have added "experimental" warning headers to the top of the documents as suggested by @sorindumitru in [0246727](https://github.com/spiffe/spiffe/pull/327/commits/0246727410ff21ad512d67ba257643fb64f01476) - let me know if any of the verbiage needs adjusting.

standards/SPIFFE_Workload_API.md
arndt-s: Let's discuss later today in SIG-SPEC. My feeling is that from a security perspective it's better to issue/release only a single credential to a workload in situations where it knows what SPIFFE-ID it requires. Issuing all and the workload just picking one seems unnecessary and when avoidable better for security.

strideynet: We discussed this during SIG-Spec, and came to the consensus that it would make sense to include this field for the WIT-SVID profile. I will adjust the spec accordingly.

strideynet: Addressed in [c2e23ee](https://github.com/spiffe/spiffe/pull/327/commits/c2e23ee464a3fe89e88eca1c7803009c17a4d26e)

standards/SPIFFE_Workload_API.md
```diff
@@ -1,5 +1,10 @@
 # The SPIFFE Workload API
 
+> [!WARNING]
```
azdagron, line 3: Is this warning still needed? It seems self referential.

standards/SPIFFE_Workload_API.md
```diff
@@ -27,18 +32,24 @@ Portable and interoperable cryptographic identity for networked workloads is per
 6\. [JWT-SVID Profile](#6-jwt-svid-profile)  
 6.1. [Profile Definition](#61-profile-definition)  
 6.2. [Profile RPCs](#62-profile-rpcs)  
-6.3. [JWT-SVID Validation](#63-jwt-svid-validation)  
+6.3. [JWT-SVID Validation](#63-jwt-svid-validation)
+7\. [WIT-SVID Profile](#7-wit-svid-profile)  
+6.1. [Profile Definition](#71-profile-definition)  
+6.2. [Profile RPCs](#72-profile-rpcs)
```
azdagron: These section numbers are off. Also, while we're here, we should add the section for the (now) section 8, Default Identity.

standards/workloadapiwitsvid.proto
```diff
@@ -0,0 +1,78 @@
+// Initial draft of WIT-SVID profile for Workload API
```
azdagron: Should this be incorporated into workloadapi.proto?

standards/SPIFFE_Workload_Endpoint.md
```diff
@@ -1,5 +1,10 @@
 # The SPIFFE Workload Endpoint
 
+> [!WARNING]
```
azdagron, line 3: Same question here. I haven't been involved in the discussions around the status of WIT-SVID support, but the relevenat changes either need to reside in a different document (marked experimental) or we should remove this notice or update it to not be self-referential (and probably relocate it to just the WIT-SVID specific sections.

standards/SPIFFE_Workload_API.md
```diff
@@ -27,18 +32,24 @@ Portable and interoperable cryptographic identity for networked workloads is per
...
+Currently, there are three profiles:
 
 - [X.509-SVID Profile](#5-x509-svid-profile)
 - [JWT-SVID Profile](#6-jwt-svid-profile)
+- [WIT-SVID Profile](#7-wit-svid-profile)
+
+The X.509-SVID and JWT-SVID profiles are mandatory and MUST be supported by SPIFFE implementations. However, operators MAY administratively disable a specific profile in their deployment.
 
-Both profiles are mandatory and MUST be supported by SPIFFE implementations. However, operators MAY administratively disable a specific profile in their deployment.
+The WIT-SVID Profile is optional and MAY be supported by SPIFFE implementations.
```
azdagron, line 52: I'm not sure if @evan2645 has had a chance to review yet, but IIRC there was a specific reason forthe language about the X.509-SVID and JWT-SVID profiles...

standards/SPIFFE_Workload_API.md
strideynet: We discussed this in SIG-Spec a while back, and I think the initial plan we settled on was to merge the WIT-related specs into a "holding branch" (`draft-wit-svid`) and iterate on these for a while.  I included this warning, since we may direct folks interested in the WIT-SVID progress to this draft branch. We'd remove this warning before merging the draft branch down to master.

Definitely open to alternative suggestions on how we handle these "experimental" status specifications.

azdagron: Ah I just noticed the target branch of the PR. This makes sense now. Thanks!

standards/workloadapiwitsvid.proto
azdagron: I'd suggest incorporating into workloadapi.proto with a clear "EXPERIMENTAL" demarcation on the WIT related portions. You'll need it that way if downstream projects are going to implement experimental support (e.g. go-spiffe).

strideynet: Agreed - that makes sense and is more in line with how we've handled the documents.

standards/SPIFFE_Workload_API.md
strideynet: Addressed in a04bb61

standards/workloadapiwitsvid.proto
strideynet: Addressed in [65dabe1](https://github.com/spiffe/spiffe/pull/327/commits/65dabe1442f8f6277943c0f93cb60cad7e828a0a)

azdagron (approved):

standards/SPIFFE_Workload_API.md
strideynet: I'm not sure I fully understand this comment - it's possible the diff is a little bit confusing here so I'll try to clarify:

What I've done here is:
- Modify the language of the existing statement that "Both profiles are mandatory" to explicitly name the JWT and X509 profiles, as, "Both" in the presence of a third profile seems confusing.
- Added an additional statement that the WIT-SVID Profile is optional.

azdagron: Let me rephrase a bit. The profile being mandatory but administratively disabled and the profile being optional are probably indistinguishable from a workload perspective (other than maybe a difference in status code). I remember us discussing the language used here in the past and there was a justification for why it is worded the way it is even though the functional result might be the same. @evan2645 might remember. I'm not sure if the WIT profile should get special treatment here or not.

standards/SPIFFE_Workload_API.md
```diff
@@ -360,6 +361,132 @@ When interfacing with legacy JWT validators, the `FetchJWTBundles` method can be
...
+message WITBundlesRequest {}
+
+// The WITBundlesResponse message carries a map of trust bundles the workload
+// should trust.
+message WITBundlesResponse {
+  // Required.
+  // WIT trust bundles belonging to trust domains that the workload should
+  // trust, keyed by the SPIFFE ID of the trust domain.
+  // Bundles are encoded in JWK set (RFC 7517) format.
+  map<string, string> bundles = 1;
```
sorindumitru, line 450: I think one thing that is missing here is the bundle format. I think it's somewhat trivial:JWK of the key with a key use of `wit-svid`, similar to the other SVID types. It would be useful to mention that somewhere.

strideynet: We discussed this in SIG-Spec last night and a couple of points came to mind:

- Today, for the JWT-SVID profile, we do not specify any restrictions or extensions beyond the standard JWKS. Therefore, the `use` field is implicitly optional - it is not required nor is it prohibited.
- Today, in the implementation of SPIRE, we do not set the `use` field to `jwt-svid` when serving the trust bundle via the Workload API.
- We do leverage the `use` field in the SPIFFE bundle provided via the federation APIs.

My gut feeling is that we should not deviate from JWT-SVID by requiring a `use` of `wit-svid`, but curious on your thoughts here?

strideynet: That being said - I did just stumble across the `kid` field in https://github.com/spiffe/go-spiffe/pull/367 and perhaps this does merit further discussion.

Today, we make no mention within this specification as to whether `kid` is mandatory for the JWKS provided by the  JWT-SVID profile - although, it is mandatory for JWT issuers within the SPIFFE bundle. However, it looks like the implementation of `spiffe-go` treats this as mandatory (the implementation of `jwtbundle` requires a kid for each keypair).

So perhaps we should take the opportunity here to make `kid` a mandatory field in the JWKS provided by the WIT-SVID Workload API profile?

sorindumitru: I think you're right about this usage of the bundle. In the Workload API we usually work with a subset of the bundle that only contains the keys for a profile. So I agree that there's no need to specify a use here.

There are cases such as federation where a bundle is exchanged in the "SPIFFE format", for that we [define](https://github.com/spiffe/spiffe/blob/main/standards/JWT-SVID.md#6-representation-in-the-spiffe-bundle) that a `key_use` is required to be able to extract the profile specific keys. That would usually go in `WIT-SVID.md`, but since we don't have that maybe we can mention something about it here?

strideynet: > That would usually go in WIT-SVID.md, but since we don't have that maybe we can mention something about it here?

I will be adding a WIT-SVID.md and making changes to the SPIFFE_Trust_Domain_and_Bundle.md in the next PR for this draft branch - so I think it makes more sense to address it there rather than duplicate that content here within the Workload API specification.

sorindumitru: Sounds good

sorindumitru (approved):
LGTM

standards/SPIFFE_Workload_API.md
```diff
@@ -360,6 +361,136 @@ When interfacing with legacy JWT validators, the `FetchJWTBundles` method can be
...
+
+The WIT-SVID Profile RPCs and associated messages are defined below. For the complete Workload API service definition, see [workloadapi.proto](workloadapi.proto).
+
+```protobuf
+service SpiffeWorkloadAPI {
+  /////////////////////////////////////////////////////////////////////////
+  // WIT-SVID Profile
+  /////////////////////////////////////////////////////////////////////////
+
+  // Fetch WIT-SVIDs for all SPIFFE identities the workload is entitled to. If
```
arndt-s, line 380: The comments here do not align with the ones from workloadapi.proto

strideynet: Should be resolved in c431c09

arndt-s (approved):

PR #340, opened 2025-05-28
SPIFFE Broker API & Endpoint

### LINKED ISSUE #313: Support for sidecar-less service meshes
The Workload API the way it is designed (and also implemented today) does not support service mesh deployments with node-level proxies such as Istio Ambient Mesh. The proxies in these deployments act on behalf of multiple workloads and would be the peers that connect to the SPIFFE Workload Endpoint. The way SPIFFE works today a proxy would only receive SVIDs representing itself but not for the workloads it acts on behalf of.

More generally, SPIFFE does not support the deployment of brokers that act on behalf of multiple workloads and sit in between the workload and the Workload API. Istio Ambient mode is one of these, but other scenarios exist too. (Kubernetes-based SPIFFE-helper, Egress API Gateways adding JWT-SVIDs to requests, etc.)

## Existing work

The [Delegated Identity API](https://github.com/spiffe/spire-api-sdk/blob/main/proto/spire/api/agent/delegatedidentity/v1/delegatedidentity.proto) of SPIRE (only SPIRE, not SPIFFE) is a dedicated API that solves this. It looks similar to the Workload API but requires certain request parameters (PID or SPIRE selectors). It is also “multi-tenanted”, returning SVIDs for all workloads (the SPIRE agent is authorized for) on a single connection.

## Proposed solution

```mermaid
sequenceDiagram
    participant Workload
    participant Broker as Broker (e.g. Proxy)
    Workload-->>Broker: Connect
alt for each workload
    Broker-->>SPIFFE Workload Endpoint:Connect
    Broker->>SPIFFE Workload Endpoint:Attest via API
    SPIFFE Workload Endpoint->>SPIFFE Workload Endpoint: Serve Workload API
    Broker->>SPIFFE Workload Endpoint:Use Workload API
    SPIFFE Workload Endpoint-->>Broker:SVID
    Broker-->>Broker:Use SVID (upgrade connection, etc.)
end
```

Add wording to the SPIFFE standard to allow attestation on API-level. In broker deployments, attestation of workloads would be based on data the broker sends once connected, rather than kernel introspection. The SPIFFE implementation validates the provided data and, if successful, starts serving the Workload API on the existing connection. The Workload API remains in the context of a single workload and follows the SPIFFE standard. The broker would open dedicated connection and attest for each workload it represents.

For the service-mesh scenario above, where a proxy representing multiple workload on the same node (like Istio Ambient mode) and identifies them using PID, the attestation part could be as simple as
```proto
// ProcessIDAttestationAPI allows you to create an attestation connection. Once successful you can use the Workload API in context of the provided PID on the connection.
service ProcessIDAttestationAPI {
	// AttestProcessID attests the channel by an ID of a process running on the same unix-based OS.
	rpc AttestProcessID(AttestProcessIDRequest) returns (AttestationResult);
}

message AttestProcessIDRequest {
	// pid is the ID of the process to attest.
	int32 pid = 1;
}

message AttestationResult {
	// spiffe_id represents all SPIFFE IDs the requested attestation context is authenticated for. It can be used as a hint in the Workload API.
	repeated string spiffe_id = 1;
	
	// expires_at is the time the attestation expires and re-attestation is required.
	google.protobuf.Timestamp expires_at = 2;
}
```
For consistency and interoperability reasons I advocate to store this API (proto) in the SPIFFE repository and allow implementations to optionally implement it.

I think this approach:

* Keeps attestation and Workload API separate concerns like it is today
* Clients continue to use the well-known Workload API and its streaming capabilities.
* Allows flexibility in the way the workload and the broker is identified/attested/authenticated. I'm thinking SPIRE plugins adding custom API-level attestation.

## Alternative solutions

Another solution is to bring the Delegated Identity API of SPIRE (or something similar) into SPIFFE. It targets above use cases and compared to the solution above is “multi-tenanted”. This comes with some optimization, particularly on bundle sync. Custom attestation cannot be easily added (maybe a good thing?).

## Discussion

* Do you know more use cases exist or are you aware of other solutions/existing work? 
* Should SPIFFE cover these deployments or should it remain an implementation detail like it is today?

arndt-s:
This PR adds a SPIFFE Broker API and Endpoint to the SPIFFE specification.

See [Google doc](https://docs.google.com/document/d/16E3B66ECE18OiY-4jilL-kqGOVITrmOSeq3yJIE-OTE/edit?tab=t.0) for design (proto is a bit outdated in it but it gives the idea)

Resolves #313.


standards/SPIFFE_Broker_API.md
```diff
@@ -0,0 +1,411 @@
...
+## 1. Introduction
+
+The SPIFFE Broker API enables Brokers to retrieve SVIDs and bundles for workloads they represent. It is served by the [SPIFFE Broker Endpoint](SPIFFE_Broker_Endpoint.md), and comprises a number of services, or *profiles*.
+
+Currently, there are two profiles:
+
+- [X.509-SVID Profile](#5-x509-svid-profile)
+- [JWT-SVID Profile](#6-jwt-svid-profile)
+
+Both profiles are mandatory and MUST be supported by SPIFFE implementations. However, operators MAY administratively disable a specific profile in their deployment.
```
strideynet, line 54: One thing we've been considering as part of the WIT-SVID work is to remove the requirement that both profiles are implemented from the Workload API specification, I think there's a few examples of SPIFFE implementations that only implement X509 and not JWT. Perhaps worth similarly dropping that requirement here as well?

arndt-s: Full disclosure - I also dropped
- the CRL on X509-SVID
- the validation RPC on JWT-SVID

Was planning to mention it in the call - but while we're talking about working in "learnings"

standards/SPIFFE_Broker_API.md
```diff
@@ -0,0 +1,411 @@
...
+
+## 4. Client and Server Behavior
+
+### 4.1 Authenticating and authorizing the Caller
+
+The SPIFFE Broker API makes use of the authentication and authorization at the [SPIFFE Broker Endpoint](./SPIFFE_Broker_Endpoint.md#5-authentication-and-authorization) to identify and authorize the caller.
+
+Implementations MUST maintain a strict allow-only policy that prevents any caller that is not authorized to leverage the SPIFFE Broker API.
+
+> Note Arndt: Do we want to talk about authorization scope? Should the standard allow Brokers to only be authorized to act on behalf of a subset of workloads or is it always global scoped?
```
kfox1111: I think that can be left as an implementation detail.

standards/SPIFFE_Broker_API.md
```diff
@@ -0,0 +1,411 @@
...
+
+The exact timing of server response messages is implementation-specific, and SHOULD be dictated by events which change the response, such as an SVID rotation. Receiving a request message from the client MUST be considered a response-generating event. In other words, the first response message of the server response stream (on a connection-by-connection basis) MUST be sent as soon as possible, without delay.
+
+Finally, implementers of SPIFFE Broker API servers should be careful about pushing updated response messages *too* rapidly. Some software may reload automatically upon receiving new information, potentially causing a period of unavailability should all instances reload at once. As a result, implementers may introduce some splay/jitter in the transmission of widespread updates.
+
+> Note Arndt: do we want to jitter responses for Brokers too, or should this behaviour not be implemented there instead? Or both?
```
kfox1111: if the agent is jittering, that would naturally jitter brokers handing it to their children too. So, that is covered already?

standards/SPIFFE_Broker_API.md
```diff
@@ -0,0 +1,411 @@
...
+    oneof reference {
+        WorkloadPIDReference pid = 1;
+    }
+
+    // Required. The audience(s) the workload intends to authenticate against.
+    repeated string audience = 2;
+
+    // Optional. The requested SPIFFE ID for the JWT-SVID. If unset, all
+    // JWT-SVIDs to which the workload is entitled are requested.
+    string spiffe_id = 3;
```
kfox1111, line 485: should there be a hint here, like in x509?

kfox1111: oh. its not a request property. nm.

standards/SPIFFE_Broker_API.md
kfox1111: Instead of not implementing the profile, it could be mandatory, but return value from it could be "not implemented" like in the workload api?

standards/SPIFFE_Broker_Endpoint.md
```diff
@@ -0,0 +1,109 @@
...
+
+In the beginnings, the on-behalf-of component was a proxy running next to the workload. Each workload would have it's own peer-proxy and together they would form a combined security boundary. In this situation, the proxy was able to leverage the SPIFFE Workload Endpoint and API to retrieve SVIDs on behalf of the workload it shares the security boundary with. This pattern is commonly known as "sidecar pattern", the proxy as a "sidecar" or "sidecar proxy" 
+
+Since this has been introduced, this pattern has evolved to locate the proxy from one per workload to one per node for security, separation of concern, maintenance and other reasons. The proxy that acts on-behalf of multiple workloads as “broker”. This pattern poses a conflict with the SPIFFE Workload API & Endpoint. The security boundary of the Broker and the workload is different, and thus the component that requests an SVID is not the workload, or a “workload specific” (sidecar) instance anymore but a broker that acts on-behalf of all the workloads running on the node. The SPIFFE of today would see the broker as the workload and not the workload the broker represents.
+
```
kfox1111, line 38: This is a valid use case, but not the only use case for this api.

The spire-ha-agent is another valid use case, but the broker is not a proxy.

standards/SPIFFE_Broker_Endpoint.md
```diff
@@ -0,0 +1,109 @@
...
+As a hardening measure against [Server Side Request Forgery](https://www.owasp.org/index.php/Server_Side_Request_Forgery) (SSRF) attacks, every client request to the SPIFFE Broker Endpoint MUST include the static gRPC metadata key `broker.spiffe.io` with a value of `true` (case sensitive). Requests not including this metadata key/value MUST be rejected by the SPIFFE Broker Endpoint (see the [Error Codes](#6-error-codes) section for more information). This prevents an attacker from exploiting an SSRF vulnerability to access the SPIFFE Broker Endpoint unless the vulnerability also gives the attacker control over outgoing gRPC metadata.
+
+### 3.1 Transport Security
+
+The SPIFFE Broker Endpoint requires transport security in the form of mutual TLS. Compared to the SPIFFE Workload Endpoint the broker and the implementation are expected to have a previously established root of trust in the form of SPIFFE bundles and X509-SVIDs. Effectively, this means that both parties need to make use of the SPIFFE Workload API as workloads first, before serving or connecting to the SPIFFE Broker Endpoint. Please see the [Authentication](#5-authentication) section for more information.
+
```
kfox1111, line 68: In the case of unix socket, is mutual tls really required? Feels like a lot of overhead when not needed?

standards/brokerapi.proto
```diff
@@ -0,0 +1,174 @@
+syntax = "proto3";
+
+import "google/protobuf/struct.proto";
+
+service SpiffeWorkloadAPI {
```
kfox1111: Should the service name be different?

standards/brokerapi.proto
```diff
@@ -0,0 +1,174 @@
...
+}
+
+// The X509SVIDResponse message carries X.509-SVIDs and related information,
+// including a list of bundles the workload may use for federating with foreign
+// trust domains.
+message X509SVIDResponse {
+    // Required. A list of X509SVID messages, each of which includes a single
+    // X.509-SVID, its private key, and the bundle for the trust domain.
+    repeated X509SVID svids = 1;
+
```
kfox1111, line 120: optional crl? we probably should follow the workload api here, or propose at the same time we remove crl from the other api.

standards/brokerapi.proto
```diff
@@ -0,0 +1,174 @@
...
+    // Required. The reference identifying the workload.
+    oneof reference {
+        WorkloadPIDReference pid = 1;
+    }
+}
+
+// The X509BundlesResponse message carries a map of trust bundles the workload 
+// should trust.
+message X509BundlesResponse {
+    // Required. CA certificate bundles belonging to trust domains that the
```
kfox1111, line 166: crl?

kfox1111:
looking really good. :)

standards/brokerapi.proto
strideynet: +1 - I think this may be a copy-paste mistake 😅

standards/SPIFFE_Broker_API.md
```diff
@@ -0,0 +1,411 @@
...
+
+If the referenced workload does not exist or is not entitled to receive any JWT bundles, then the server SHOULD respond with the "FailedPrecondition" gRPC status code. See [Section 7](#7-workload-reference-error) for details. Under such a case, the Broker MAY attempt to reconnect with another call to the `FetchJWTBundles` RPC after a backoff.
+
+As mentioned in [Stream Responses](#43-stream-responses), each `JWTBundlesResponse` response contains the complete set of authorized JWT bundles of the workload at that point in time. As such, if the server redacts bundles from a subsequent response that was in context of the referenced workload the Broker SHOULD cease using the redacted bundles. This includes situations where the server returns a Permission denied, where the Broker is expected to drop all previous received bundles.
+
+# 7. Workload Reference Error
```
strideynet: Do you have any examples of what this would look like in Go (or similar) on the implementation? I'm usually familiar with `google.rpc.Status` being used to convey additional error details for gRPC and it's not too clear how this fits in with that or replaces it.

standards/SPIFFE_Broker_API.md
arndt-s: SPIFFE SIG-Spec call today agrees with this.

standards/SPIFFE_Broker_Endpoint.md
arndt-s: Sharing some of my thoughts to go down this way: 
- With the strong authorization requirements I believe using mTLS gives us the benefit that we don't have to perform attestation, find potentially multiple identities and perform the authz check for each. It's much more explicit.
- Brokers get the guarantee they are talking to a valid SPIFFE server with the server certificate.
- We keep the "unauthenticated" APIs to a minimum.

I also think that it decouples attestation from the Broker API but that's only partially true as the Broker server still needs to perform attestation for the PIDs in the requests ..

standards/brokerapi.proto
arndt-s: We are in the progress of doing this, it stalled a bit though.
https://github.com/spiffe/spiffe/pull/333

standards/SPIFFE_Broker_Endpoint.md
arndt-s: @kfox1111 very interesting! I added a sentence to capture this use case, can you take a look please and let me know if I understand it correctly?

standards/brokerapi.proto
arndt-s: Good catch, thank you!

standards/SPIFFE_Broker_Endpoint.md
kfox1111: Strong attestation can be done via unix socket without mtls. That is how the regular spiffe workload api works today.

your right, in that it simplifies the code on the one end to not have to lookup the details. but then the other end has to do it. So, maybe the same amount of work either way?

The existing spire-agent implementation does it by unix socket identity resolving within the agent to the svid. So the code to check is already in the agent today too. So would be easier to adapt if the protocol was a little more simlilar.

kfox1111: oh. and I'm guessing there will be fewer agents written, then brokers. So simplifying what the broker needs to do might have some advantages.

apexlnc:
are there plans to continue work on this proposal? with istio ambient mode reaching GA, spire support is needed. 

https://github.com/istio/istio/issues/42339#issuecomment-3382357524

arndt-s:
Hi @apexlnc,

this topic was discussed at the [Istio spec call](https://docs.google.com/document/d/1wsa06GGiq1LEGwhkiPP0FKIZJqdAiue-VeBonWAzAyk/edit?tab=t.0#heading=h.o8pz6aqnzzgk) last week. The Istio community has yet to decide if they want to support a SPIFFE API at ztunnel or other flavors such as Envoy xDS. Until this decision is made this work here is put on hold.

Feel free to join today and express your opinion in this topic.

kfox1111:
1. There are some of us that are avoiding using istio (ambient) right now because it does not support spiffe. So they really should take that into consideration.
2. I think we should have this functionality here, regardless if istio will support it or not.

arndt-s:
@kfox1111 I'm not a bit hesitant adding this API into the standard if we have no client in sight that's picking it up. (In case Istio decides against it)

strideynet:
I agree with Arndt's sentiment here - if we don't have a clear adopter in mind, I would be hesitant to introduce the new specification. It'll be difficult for us to validate the specification is fit for purpose.

kfox1111:
I'd use it for making spire-ha-agent -> spiffe-ha-agent.

There's also a chicken and the egg issue around it. We may never get users if we dont produce it in the first place. As is, istio ambient has been delaying as "we dont have spiffe support" If we had it, then it would shine light on the issue and prevent that from being used as a reason for not supporting it.

strideynet:
> I'd use it for making spire-ha-agent -> spiffe-ha-agent.

Is there some documentation/design somewhere for this? It'd be great to explore this in SIG-Spec

kfox1111:
Just whats in https://github.com/spiffe/spire-ha-agent

standards/brokerapi.proto
```diff
@@ -0,0 +1,174 @@
...
+    // SPIFFE ID of the trust domain. As this information changes, subsequent
+    // messages will be streamed from the server.
+    rpc FetchJWTBundles(JWTBundlesRequest) returns (stream JWTBundlesResponse);
+}
+
+// The WorkloadPIDReference message conveys a process id reference of a workload 
+// running in the same environment.
+message WorkloadPIDReference {
+    // Required. The process id of the workload.
+    string pid = 1;
```
MikeZappa87: Why a string?

arndt-s: Good point. This should be an int32.

standards/brokerapi.proto
```diff
@@ -0,0 +1,174 @@
...
+
+    // Fetches the JWT bundles, formatted as JWKS documents, keyed by the
+    // SPIFFE ID of the trust domain. As this information changes, subsequent
+    // messages will be streamed from the server.
+    rpc FetchJWTBundles(JWTBundlesRequest) returns (stream JWTBundlesResponse);
+}
+
+// The WorkloadPIDReference message conveys a process id reference of a workload 
+// running in the same environment.
+message WorkloadPIDReference {
```
MikeZappa87, line 47: In K8s when containerd is used as the container runtime, a pod has a 'pause' container and then the workload containers. We can have 2 or more pids for example (I don't think crio has a pause container anymore) can it be clarified which pid would be used in this case? I know in the spire delegated identity, if you pass the pause pid it fails attestation. It would be great to have this explicitly stated? Maybe I just need to go back and read the original documents related to that effort and missed it.

Jeansen: CRI-O still utilizes a pause container.

Jeansen:
> I'd use it for making spire-ha-agent -> spiffe-ha-agent.
> 
> There's also a chicken and the egg issue around it. We may never get users if we dont produce it in the first place. As is, istio ambient has been delaying as "we dont have spiffe support" If we had it, then it would shine light on the issue and prevent that from being used as a reason for not supporting it.

I agree. We already use SPIRE with Istio's sidecar mode. I am eager to replace the sidecars with ambient ... but without the SPIRE/SPIFFE support, that is actually the only pain point why we did not migrate to ambient ;-(

kfox1111:
Same. I'm hesitant to deploy ambient without spire support too.

Jeansen:
I actually wonder a bit. Just stumpled over: https://www.solo.io/blog/kubernetes-identity-the-right-way-with-spire-and-ambient

MikeZappa87:
> I actually wonder a bit. Just stumpled over: https://www.solo.io/blog/kubernetes-identity-the-right-way-with-spire-and-ambient

Solo has spire integration with the enterprise version of ztunnel. I am currently putting together a github issue in ztunnel to start a channel to get my version of ztunnel that supports Spire upstreamed for the broader community.

Jeansen:
Mhm...too bad SOLO did not contribute it to open source community. @MikeZappa87 if there is something to test, let me know. I'd be eager to try this out.

kamalmarhubi:
As @kfox1111 said, there seems to be a chicken-and-egg problem here.

[Further up](https://github.com/spiffe/spiffe/pull/340#issuecomment-3407190810) in this issue
> @kfox1111 I'm not a bit hesitant adding this API into the standard if we have no client in sight that's picking it up. (In case Istio decides against it)

An Istio maintainer saying [a few months ago](https://github.com/istio/istio/issues/42339#issuecomment-2913547973):
> That would be ok with me; moving from e.g. experimental -> alpha would be gated on the spiffe API

Seems like the kind of thing where a relatively short synchronous conversation between two parties could sort out if both projects are willing to move forward.

(Also posting on the reciprocal issue on istio.)

arndt-s:
@kamalmarhubi I'm working together with the Istio community to get this work finished. There is a [weekly community call](https://github.com/istio/community/blob/master/WORKING-GROUPS.md) where this topic is discussed.

apexlnc:
@arndt-s appreciate you taking this forward. are there meeting notes and/or next steps on this?

arndt-s:
The latest update (approx. 3 weeks ago) is that the Istio community is looking at various alternative approaches before making a decision. One of them being leveraging CAP_SYS_ADMIN to make the existing Workload API incl. PID attestation work. Another alternative is using Envoy SDS (not the current variant SPIRE has implemented but an alternative to it where PID is passed as resource names IIRC)

Please feel free to reach out or comment here if you have perspectives. The notes/agenda from the Istio Working Group meeting can be found here: https://docs.google.com/document/d/1wsa06GGiq1LEGwhkiPP0FKIZJqdAiue-VeBonWAzAyk/edit?tab=t.0#heading=h.o8pz6aqnzzgk

MikeZappa87:
> The latest update (approx. 3 weeks ago) is that the Istio community is looking at various alternative approaches before making a decision. One of them being leveraging CAP_SYS_ADMIN to make the existing Workload API incl. PID attestation work. Another alternative is using Envoy SDS (not the current variant SPIRE has implemented but an alternative to it where PID is passed as resource names IIRC)
> 
> Please feel free to reach out or comment here if you have perspectives. The notes/agenda from the Istio Working Group meeting can be found here: https://docs.google.com/document/d/1wsa06GGiq1LEGwhkiPP0FKIZJqdAiue-VeBonWAzAyk/edit?tab=t.0#heading=h.o8pz6aqnzzgk

We have this working however we required to use the host pid namespace (hostPid: true)

arndt-s:
@MikeZappa87 Which option do you have working? The one that uses the existing WorkloadAPI or via SDS?

standards/brokerapi.proto
arndt-s: Removed it from here to keep the changes concise and focused.

standards/SPIFFE_Broker_API.md
arndt-s: Agreed. I think it's reasonable enough to mention jitter here. Whether it's then used is an implementation detail.

standards/brokerapi.proto
arndt-s, line 1: Note to myself: Consider SPIRE extensions

standards/SPIFFE_Broker_API.md
```diff
@@ -0,0 +1,385 @@
+# The SPIFFE Broker API
+
+## Status of this Memo
+
+This document specifies an identity API standard for the internet community, and requests discussion and suggestions for improvements. Distribution of this document is unlimited.
+
+## Abstract
+
+Brokers are trusted infrastructure components that can act on-behalf-of workloads. This API enables them to retrieve SVIDs and bundles of workloads it represents by referencing the workloads by its process ID.
```
jsnctl: "bundles of workloads" kind of reads like "bundles, which contain workloads". Amendment suggested
```suggestion
Brokers are trusted infrastructure components that can act on-behalf-of workloads. This API enables them to retrieve the SVIDs and trust bundles of workloads it represents by referencing the workloads by its process ID.
```

standards/SPIFFE_Broker_API.md
```diff
@@ -0,0 +1,385 @@
...
+
+If the referenced workload does not exist or is not entitled to receive any X509-SVIDs, then the server MUST respond with an appropriate gRPC status code as specified in [Section 4.8](#48-workload-references-and-svid-entitlements). Under such a case, the Broker MAY attempt to reconnect with another call to the `FetchX509SVID` RPC after a backoff.
+
+As mentioned in [Stream Responses](#43-stream-responses), each `X509SVIDResponse` message returned on the `FetchX509SVID` stream contains the complete set of authorized SVIDs and bundles of the workload at that point in time. As such, if the server redacts SVIDs from a subsequent response that was in context of the referenced workload the Broker SHOULD cease using the redacted SVIDS. This includes situations where the server returns a Permission denied, where the Broker is expected to drop all previous received SVIDs and bundles.
```
jsnctl:
```suggestion
As mentioned in [Stream Responses](#43-stream-responses), each `X509SVIDResponse` message returned on the `FetchX509SVID` stream contains the complete set of authorized SVIDs and bundles of the workload at that point in time. As such, if the server redacts SVIDs from a subsequent response that was in context of the referenced workload the Broker SHOULD cease using the redacted SVIDs. This includes situations where the server returns a Permission denied, where the Broker is expected to drop all previous received SVIDs and bundles.
```

standards/SPIFFE_Broker_API.md
```diff
@@ -0,0 +1,385 @@
...
+
+As mentioned in [Stream Responses](#43-stream-responses), each `X509BundleResponse` response contains the complete set of authorized X.509 bundles of the workload at that point in time. As such, if the server redacts bundles from a subsequent response that was in context of the referenced workload the Broker SHOULD cease using the bundles. This includes situations where the server returns a Permission denied, where the Broker is expected to drop all previous received bundles.
+
+### 5.3 Default Identity
+
+Servers and clients of Broker API are expected to follow the behavior as defined in section 5.3 of the [SPIFFE Workload API](./SPIFFE_Workload_API.md#53-default-identity) when multiple identities are returned.
+
+The Broker is expected to use the default identity in situations where it cannot determine the expected usage and is not able to deliver multiple identities to the workload. When it is able to deliver multiple identities to workloads it MUST identify these identities with their corresponding `hint` as delivered by the SPIFFE Broker API and clients MUST be able to use these identifier to identify identities.
```
jsnctl:
```suggestion
The Broker is expected to use the default identity in situations where it cannot determine the expected usage and is not able to deliver multiple identities to the workload. When it is able to deliver multiple identities to workloads it MUST identify these identities with their corresponding `hint` as delivered by the SPIFFE Broker API and clients MUST be able to use these identifiers to identify identities.
```

standards/SPIFFE_Broker_API.md
```diff
@@ -0,0 +1,385 @@
...
+
+The exact timing of server response messages is implementation-specific, and SHOULD be dictated by events which change the response, such as an SVID rotation. Receiving a request message from the client MUST be considered a response-generating event. In other words, the first response message of the server response stream (on a connection-by-connection basis) MUST be sent as soon as possible, without delay.
+
+Finally, implementers of SPIFFE Broker API servers should be careful about pushing updated response messages *too* rapidly. Some software may reload automatically upon receiving new information, potentially causing a period of unavailability should all instances reload at once. As a result, implementers may introduce some splay/jitter in the transmission of widespread updates.
+
+### 4.5 Default Values and Redacted Information
+
+Client and servers MUST implement the same default values behavior as described in section 4.4 of the [SPIFFE Workload API](./SPIFFE_Workload_API.md).
```
jsnctl:
```suggestion
Clients and servers MUST implement the same default values behavior as described in section 4.4 of the [SPIFFE Workload API](./SPIFFE_Workload_API.md).
```

standards/SPIFFE_Broker_API.md
```diff
@@ -0,0 +1,385 @@
...
+Servers SHOULD include `google.rpc.ErrorInfo` details with the error response containing:
+- `reason`: One of the error reasons specified in the table above
+- `domain`: "spiffe.io"
+- `metadata`: Optional contextual information such as the PID value or other debugging details
+
+Clients MAY inspect ErrorInfo details for structured error information but MUST handle cases where only the status code is available.
+
+### 4.9 Workload Lifecycle
+
+Both, server and client MUST monitor the state of the workload and ensure that no operations are performed beyond the lifetime of the workload. For instance, the server MUST not send responses to the client once the workload has stopped. Clients on the other hand MUST drop all the data received for the workload, removing it from file systems or other locations it potentially have stored it in addition.
```
jsnctl:
```suggestion
Both server and client MUST monitor the state of the workload and ensure that no operations are performed beyond the lifetime of the workload. For instance, the server MUST not send responses to the client once the workload has stopped. Clients on the other hand MUST drop all the data received for the workload, removing it from file systems or other locations it potentially have stored it in addition.
```

standards/SPIFFE_Broker_Endpoint.md
```diff
@@ -0,0 +1,111 @@
...
+## 5. Authentication and Authorization
+
+The SPIFFE Broker Endpoint requires mutual authentication in the form of Mutual Transport Layer Security as defined in [RFC 5246](https://www.rfc-editor.org/rfc/rfc5246.txt) (TLS 1.2) and [RFC 8446](https://www.rfc-editor.org/rfc/rfc8446.txt) (TLS 1.3) in the form of X509-SVIDs. Prior to the mutual authentication both parties need to retrieve the SVID and bundle, for instance by using the SPIFFE Workload API.
+
+The nature of the SPIFFE Broker Endpoint requires strong access control and only authorized callers must be allowed to connect. Controlling access only by accessibility is highly discouraged and implementors should restrict both, only making the endpoint accessible to allowed components and implementing strict access control on the endpoint. It is reccomended to base access control on the SPIFFE-ID of the client. Given the broad capabilities the SPIFFE Broker Endpoint offers to authorized clients the policy is reccommended to be static and based of configuration.
```
jsnctl: Both "implementer" and "implementor" are used in the doc - I think the technical standard is to use -er but I'm not sure

jsnctl (commented):
Very quick pass on various typos and tidying bits

standards/brokerapi.proto
arndt-s: Opened this up again as #333 didn't progress and I don't want to remove a field indirectly but "not including it into a new API".

Let's discuss next SIG-SPEC whether the Broker API should include it or not.

standards/SPIFFE_Broker_Endpoint.md
arndt-s: Good catch, fixed (using implementer)

standards/brokerapi.proto
arndt-s: Background: SPIRE allows to plug in custom workload attestors. The underlying API is the following: https://github.com/spiffe/spire-plugin-sdk/blob/main/proto/spire/plugin/agent/workloadattestor/v1/workloadattestor.proto
```
message AttestRequest {
    // Required. The process ID of the workload to attest.
    int32 pid = 1;
}
```

How would this work with the new API allowing to pass in references outside of the PID?

standards/brokerapi.proto
```diff
@@ -0,0 +1,164 @@
...
+    rpc FetchX509Bundles(X509BundlesRequest) returns (stream X509BundlesResponse);
+
+    /////////////////////////////////////////////////////////////////////////
+    // JWT-SVID Profile
+    /////////////////////////////////////////////////////////////////////////
+
+    // Fetch JWT-SVIDs for all SPIFFE identities the referenced workload is 
+    // entitled to, for the requested audience. If an optional SPIFFE ID is 
+    // requested, only the JWT-SVID for that SPIFFE ID is returned.
+    rpc FetchJWTSVID(JWTSVIDRequest) returns (JWTSVIDResponse);
```
arndt-s: I noticed that SPIRE Delegated Identity API doesn't return JWTSVID stream. I think we should
- Differentiate naming, so it's easier to add JWTSVID streaming later
- Offer unary requests for unary use cases (I noticed that ztunnel doesn't use the streams anyway)

standards/brokerapi.proto
```diff
@@ -0,0 +1,166 @@
...
+message WorkloadReference {
+    // Required. The reference to the workload.
+    google.protobuf.Any reference = 1;
+}
+
+// The WorkloadPIDReference message conveys a process id reference of a workload
+// running in the same environment.
+message WorkloadPIDReference {
+    // Required. The process id of the workload. MUST be a positive integer.
+    int32 pid = 1;
```
sorindumitru, line 52: Can't this just be `uint32`?
```suggestion
    uint32 pid = 1;
```

standards/SPIFFE_Broker_API.md
```diff
@@ -0,0 +1,465 @@
...
+
+The SPIFFE Broker API requires a mechanism to uniquely identify and reference workloads running within the same environment as the Broker. This reference system enables Brokers to request identity materials on behalf of specific workloads while maintaining proper isolation and security boundaries.
+
+Workload references MUST satisfy the following requirements:
+
+* Uniqueness: Each reference must uniquely identify a single workload within the local environment at any given time
+* Local Scope: References are valid only within the local execution environment and MUST NOT be used across network boundaries or different nodes
+* Dereferenceability: The server must be able to verify that the referenced workload exists and is accessible for identity operations
+
+### 3.1.1 Multiple Reference Support
```
arndt-s: We should prevent multiple references of the same type IMO.

standards/SPIFFE_Broker_API.md
```diff
@@ -0,0 +1,491 @@
...
+
+All request messages in the SPIFFE Broker API accept one or more references to identify a workload. When multiple references are provided, they MUST all resolve to the same workload. This provides stronger security guarantees by requiring multiple proofs of workload identity.
```
sorindumitru: I think it might be hard to determine if it resolves to the same workload. Should this say to the same SPIFFE ID? That's effectively the same from our point of view, but it's clearer what we mean.

standards/brokerapi.proto
mikebrow: last I checked pause container was default on but could be disabled in crio

standards/SPIFFE_Broker_API.md
```diff
@@ -0,0 +1,491 @@
...
+    // messages will be streamed from the server.
+    rpc SubscribeToJWTBundles(SubscribeToJWTBundlesRequest) returns (stream SubscribeToJWTBundlesResponse);
+
+    // ... RPCs for other profiles ...
+}
+
+// The WorkloadReference message represents a single reference to a workload.
+message WorkloadReference {
+    // Required. The reference to the workload.
+    google.protobuf.Any reference = 1;
```
sorindumitru, line 462: I was asking for more customisation here to support the fact that SPIRE allows for custom attestation methods. I don't think that's actually needed. If we have the pid reference we can use any of the existing attestation methods, it's just that workloads won't be able to use any custom references without first defining them in the standard.

I think having this be a OneOf (or something like that) will end up being better for the standard and guarantee better interoperability between different implementations.

matheuscscp:
Hello! As discussed in person with @arndt-s at the SPIFFE KubeCon booth, the Flux project is interested in consuming this API for issuing short-lived creds for both the Flux controllers (Kubernetes pods) and Flux objects (Kubernetes objects), so we would love to see support for issuing SVIDs for objects added to this API :pray:

MikeZappa87:
> Hello! As discussed in person with @arndt-s at the SPIFFE KubeCon booth, the Flux project is interested in consuming this API for issuing short-lived creds for both the Flux controllers (Kubernetes pods) and Flux objects (Kubernetes objects), so we would love to see support for issuing SVIDs for objects added to this API 🙏

@arndt-s what can we do to get this merged? Can we set a timeline and merge? Then just capture any tech debt items? I fear this PR will just sit in limbo forever.

kfox1111:
I was looking at adding Federation support to the spire-ha-agent, but it would be significantly easier if I could use this API. Also looking for a timeline where this may be merged.

strideynet:
re: Timeline.

We should probably discuss this in SIG-Spec as to determine what the process here will be. I don't think it's as simple as "merge and address tech-debt later" since we likely want to avoid making breaking changes to specs that have been released.

We don't actually have a defined mechanism for graduating specs up through stability levels at the moment, but with the WIT-SVID we have been trialling treating the branch as an "alpha" of the specification with reduced stability guarantees.

In my own opinion, I would want to see this spec implemented in SPIRE and consumed by an implementation before considering it "stable". As for how we move to this point, I can see a few options:

1. Final call for review on this PR, SIG-Spec then denotes this as "alpha/beta" with some degree of stability pending running code. When there is satisfaction that the API proposed here is fit for use as proven by real usage, then we merge down and consider this a stable part of the specification.
2. Final call for review on this PR, we add an "unstable" header to the specification, and merge. At a later date, PR is used to remove the "unstable" header.

SPIFFE likely doesn't need as much procedure as IETF, but this does get me thinking as to if we should define a light set of process and criteria around spec stability and the graduation process?

I'm definitely keen to see us give this PR some kind of marker of "stableish" so we can unblock implementation in SPIRE/consumers. I'll put it in my calendar to give this PR a final sweep of review from myself.

@arndt-s - do you think it would be valuable to table some meta discussion re this spec and the release process at an upcoming SIG-SPEC?

matheuscscp:
I would really love to implement a proof of concept with Flux to show that the current state of this spec can address the following use cases:

* The flux-system namespace has the PSA annotation that prevents host-level things, e.g. SPIRE's current hostPath volume for the admin socket of the Delegated Identity API. TCP support is very important for us so that we don't need to lower the security of this namespace, which is on par with kube-system in terms of trust requirements. (~~Question: Will the TCP support also entail encryption of the communication or will it be plaintext like other workload identity daemons e.g. gke-metadata-server?~~ Edit: I see this PR says the broker must use X509-SVID for mTLS)
* Flux is always the first component installed in a cluster. Our most important operation is bootstrapping a cluster: Flux gets installed imperatively or via Terraform, and then takes over upgrading itself and installing everything else in the cluster, including SPIRE. But we need to mount the CSI driver in the Flux controllers so they can use SPIRE to issue SVIDs for the Flux Kubernetes objects. Someone told me the pods can come up healthy if the CSI driver YAML is there even if the driver does not exist yet, and that when it comes up later (after Flux installs SPIRE) it will work. This is a bold claim and I need to test this myself before relying on it. I'm not sure however if there's anything that can/should be done in this specific proposal to improve this UX, though, but from our perspective of planning the whole Flux <---> SPIFFE/SPIRE integration we need to test everything.

I'm very keen on working on this proof of concept with utmost urgency to get this done fast and giving feedback ASAP. If there's any experimental code pointers that would help me hack a proof of concept so I can work as fast as possible to help this proposal I'd love it. During this KubeCon I made a proof of concept using SPIRE's admin socket very fast in one of the nights.

I'm also very keen on joining the sig spec meetings, is there a website with the pointers for joining the meeting?

Thanks in advance for all the work and help here, we (the Flux maintainers) are very excited with this proposal and would love to help make it very solid.

arndt-s:
@matheuscscp sig-spec is happening every Thursday. [Here are details](https://github.com/spiffe/spiffe/blob/main/community/sig-spec/README.md), including a link to the calendar with join instructions.

@strideynet agreed, let's have this discussion next week at sig-spec.

@MikeZappa87 I will get back to you next week. Things have been busy the last weeks but I'm hoping to carve out some time to get the SPIRE changes done. I'm already happy with the overall structure but I'm still missing to change some of the attestation logic which currently is processID only and now needs to handle pod UIDs. This also involves an option to query the API server where before it was only kubelet.

matheuscscp:
@arndt-s Are you planning to have a working PR for SPIRE before this PR gets merged? If so, can you point me the branch so I can keep an eye on it and possibly also test it with Flux?

matheuscscp:
@arndt-s After what we discussed in the last meeting, this is my suggestion for the runtime to verify if the broker workload calling the API can act on behalf of the Kubernetes object it is requesting SVIDs for:

Proof of concept here: https://github.com/matheuscscp/spire-broker-api

matheuscscp:
Then the SPIFFE ID should be constructed from those input fields:

```go
				Group:     "example.io",
				Resource:  "widgets",
				Namespace: "object-namespace",
				Name:      "object-name",
```

matheuscscp:
@arndt-s I reached my conclusion for the best approach for k8s object refs. and SPIFFE ID. I added also a few `Fetch*` variants to the same PR, as Flux cannot consume SVID streams, it needs to fetch them and keep them cached during their attested lifetime. PR: https://github.com/arndt-s/spiffe/pull/1

standards/SPIFFE_Broker_API.md
```diff
@@ -0,0 +1,491 @@
+# The SPIFFE Broker API
+
+## Status of this Memo
+
+This document specifies an identity API standard for the internet community, and requests discussion and suggestions for improvements. Distribution of this document is unlimited.
+
+## Abstract
+
+Brokers are trusted infrastructure components that can act on-behalf-of workloads. This API enables them to retrieve the SVIDs and trust bundles of workloads it represents by referencing the workloads by its process ID.
```
amartinezfayo: Should we update this section to not mention exclusively "process ID" ?
Maybe something like "by referencing the workloads through one or more workload references".

standards/SPIFFE_Broker_API.md
```diff
@@ -0,0 +1,491 @@
...
+Servers SHOULD include `google.rpc.ErrorInfo` details with the error response containing:
+- `reason`: One of the error reasons specified in the table above
+- `domain`: "spiffe.io"
+- `metadata`: Optional contextual information such as the PID value or other debugging details
+
+Clients MAY inspect ErrorInfo details for structured error information but MUST handle cases where only the status code is available.
+
+### 4.9 Workload Lifecycle
+
+Both server and client MUST monitor the state of the workload and ensure that no operations are performed beyond the lifetime of the workload. For instance, the server MUST not send responses to the client once the workload has stopped. Clients on the other hand MUST drop all the data received for the workload, removing it from file systems or other locations it potentially have stored it in addition.
```
amartinezfayo: Should we be clearer about how a server detects end-of-lifetime? That signal is reference-type-specific (process exit for PID, pod deletion for K8s pod UID, extension-defined otherwise), and without explicit guidance a server accepting multiple reference types on the same connection has to figure it out per type.
Or maybe this was written this way to let implementers decide?

standards/SPIFFE_Broker_API.md
```diff
@@ -0,0 +1,491 @@
...
+4.5. [Mandatory Fields](#45-mandatory-fields)  
+4.6. [Federated Bundles](#46-federated-bundles)  
+5\. [X.509-SVID Profile](#5-x509-svid-profile)  
+5.1. [Profile Definition](#51-profile-definition)  
+5.2. [Profile RPCs](#52-profile-rpcs)  
+5.3. [Default Identity](#53-default-identity)  
+6\. [JWT-SVID Profile](#6-jwt-svid-profile)  
+6.1. [Profile Definition](#61-profile-definition)  
+6.2. [Profile RPCs](#62-profile-rpcs)  
+6.3. [JWT-SVID Validation](#63-jwt-svid-validation)
```
amartinezfayo: I don't think we will have a "JWT-SVID Validation" section in this spec, right?

standards/SPIFFE_Broker_Endpoint.md
```diff
@@ -0,0 +1,111 @@
...
+
+## 4. Locating the Endpoint
+
+Clients may be explicitly configured with the socket location, or may utilize the well-known environment variable `SPIFFE_BROKER_SOCKET`. If not explicitly configured, conforming clients MUST fall back to the environment variable.
+
+The value of the `SPIFFE_BROKER_SOCKET` environment variable is structured as an [RFC 3986](https://www.ietf.org/rfc/rfc3986.txt) URI. The scheme MUST be set to either `unix` or `tcp`, which indicates that the endpoint is served over a Unix Domain Socket or a TCP listen socket, respectively.
+
+If the scheme is set to `unix`, then the authority component MUST NOT be set, and the path component MUST be set to the absolute path of the SPIFFE Workload Endpoint Unix Domain Socket (e.g. `unix:///path/to/endpoint.sock`). The scheme and path components are mandatory, and no other component may be set.
```
amartinezfayo:
```suggestion
If the scheme is set to `unix`, then the authority component MUST NOT be set, and the path component MUST be set to the absolute path of the SPIFFE Broker Endpoint Unix Domain Socket (e.g. `unix:///path/to/endpoint.sock`). The scheme and path components are mandatory, and no other component may be set.
```

standards/SPIFFE_Broker_Endpoint.md
```diff
@@ -0,0 +1,111 @@
...
+
+If the scheme is set to `tcp`, then the host component of the authority MUST be set to an IP address, and the port component of the authority MUST be set to the TCP port number of the SPIFFE Workload Endpoint TCP listen socket. The scheme, host, and port components are mandatory, and no other component may be set. As an example, `tcp://127.0.0.1:8000` is valid, and `tcp://127.0.0.1:8000/foo` is not.
```
amartinezfayo:
```suggestion
If the scheme is set to `tcp`, then the host component of the authority MUST be set to an IP address, and the port component of the authority MUST be set to the TCP port number of the SPIFFE Broker Endpoint TCP listen socket. The scheme, host, and port components are mandatory, and no other component may be set. As an example, `tcp://127.0.0.1:8000` is valid, and `tcp://127.0.0.1:8000/foo` is not.
```

standards/SPIFFE_Broker_API.md
```diff
@@ -0,0 +1,491 @@
...
+
+Implementations extending the reference types SHOULD document their extensions, including the fully-qualified type name used in the Any field, to avoid conflicts with other implementations. Servers that receive a reference type they do not recognize MUST reject the request with an InvalidArgument status.
+
+## 4. Client and Server Behavior
+
+### 4.1 Authenticating and authorizing the Caller
+
+The SPIFFE Broker API makes use of the authentication and authorization at the [SPIFFE Broker Endpoint](./SPIFFE_Broker_Endpoint.md#5-authentication-and-authorization) to identify and authorize the caller.
+
+Implementations MUST maintain a strict allow-only policy that prevents any caller that is not authorized to leverage the SPIFFE Broker API.
```
amartinezfayo, line 224: I wonder if we should note that this authorization MAY also be scoped per reference type per caller? spiffe/spire#6915 adds an `AllowedReferenceTypesByCaller` map that restricts which `WorkloadReference` types each authenticated broker may use, as an example of how this is enforced in an implementation.

matheuscscp:
@arndt-s @strideynet Can we book a calender date to merge this? We have a solid implementation here (reviewed by 4 people, comments addressed, e2e tests, docs, CI green):

https://github.com/spiffe/spire/pull/6915

standards/SPIFFE_Broker_API.md
arndt-s: Removed

standards/SPIFFE_Broker_API.md
arndt-s: Good point.

I've added the following section:

> Implementations MAY enforce more fine-grained access control by inspecting data carried in the request. For example, an implementation MAY restrict a given caller to a subset of reference types (such as only `WorkloadPIDReference`, or only `KubernetesObjectReference` against a specific namespace), to a subset of audiences in `FetchJWTSVID`, or to a subset of requested SPIFFE IDs. The exact policy model is implementation-defined.

standards/SPIFFE_Broker_API.md
arndt-s: This is outdated. We only allow a single reference now.

standards/SPIFFE_Broker_API.md
arndt-s: This section got removed.

standards/brokerapi.proto
arndt-s: The `crl` is present in the Broker API now and we can look at it in the context of #333.

standards/SPIFFE_Broker_API.md
arndt-s: added in b81983a

standards/SPIFFE_Broker_API.md
arndt-s: @sorindumitru is this comment still valid today or can I resolve it?

standards/brokerapi.proto
arndt-s: @sorindumitru I believe int32 fits nicer into how pid is represented today. While I agree that uint32 is the better presentation it will most likely require mapping within client & servers.

standards/SPIFFE_Broker_API.md
arndt-s: I'm a bit undecided here. I'm inclining to make it implementation specific but I can see benefit making this more specific, but it also complicates things.

arndt-s: Actually, I've added a clarification that only the ones for the channel is to be dropped, not all. Does that solve the comment?

standards/brokerapi.proto
sorindumitru: Yes, that's ok with me. And looking now at pid_t, at least on linux it's defined as being signed, so `int32` is the right type

standards/SPIFFE_Broker_API.md
```diff
@@ -0,0 +1,642 @@
...
+- The `plural` and `group` components map directly onto the `Resource` and
+  `Group` fields of a `SubjectAccessReview` request without any string parsing.
+- It avoids ambiguity when a custom resource's plural name happens to contain
+  a dot (e.g., `myresource.example` in the `com` group vs. `myresource` in the
+  `example.com` group), which the `<plural>.<group>` shorthand cannot
+  disambiguate.
+
+Note that SPIFFE does not specify how runtimes should attest workloads referenced
+by a `KubernetesObjectReference`, but this design facilitates implementations that
+choose to attest via the recommended `SubjectAccessReview` API in Kubernetes.
```
sorindumitru: I'm not sure how this fits in. The `SubjectAccessReview` checks if a user can perform a certain action, but what's the user is in this case?

standards/SPIFFE_Broker_API.md
```diff
@@ -0,0 +1,642 @@
...
+    type_url: "type.googleapis.com/KubernetesObjectReference"
+    value: <packed KubernetesObjectReference {
+      type: { plural: "nodes", group: "core" }
+      key: { name: "ip-10-0-1-42.ec2.internal" }
+    }>
+  }
+}
+```
+
+### 3.1.4 Extensibility
```
sorindumitru, line 210: We also have section 2 on extensibility, which  says that the API is not extensible. I think that only applies to RPCs, but it could be confusing to have both sections in here.

standards/brokerapi.proto
```diff
@@ -0,0 +1,217 @@
...
+// running in the same environment.
+message WorkloadPIDReference {
+    // Required. The process id of the workload. MUST be a positive integer.
+    // For workloads running inside container runtimes that maintain a separate
+    // sandbox (e.g. the `pause` container in containerd or CRI-O), this MUST
+    // be the PID of the workload's own process, not the sandbox PID.
+    int32 pid = 1;
+}
+
+// The KubernetesObjectType message identifies the type of a Kubernetes object
```
strideynet, line 55: Nit: Would it be more accurate in Kubernetes parlance to refer to this as Kind rather than Type? It's been a while since I worked with Kube APIs.

standards/brokerapi.proto
```diff
@@ -0,0 +1,217 @@
...
+// happen to contain dots (e.g., the plural `myresource.example` in the `com`
+// group vs. `myresource` in the `example.com` group) that arises when the
+// `<plural>.<group>` shorthand is parsed as a single string.
+message KubernetesObjectType {
+    // Required. The plural name of the Kubernetes resource, as registered with
+    // the Kubernetes API server. For example, `pods`, `deployments`, or the
+    // plural form declared by a `CustomResourceDefinition`.
+    string plural = 1;
+
+    // Required. The API group of the Kubernetes resource. For non-core
```
strideynet, line 66: Nit: Should we explicit state somewhere that we don't care about version?

standards/SPIFFE_Broker_API.md
strideynet: If I recall prior discussions, one concern with `oneof` is that we'd effectively need to act as the registry/coordinate field numbers between parties - whereas for Any it's a little bit more natural.

standards/SPIFFE_Broker_API.md
```diff
@@ -0,0 +1,642 @@
...
+
+    // Optional. CA certificate bundles belonging to foreign trust domains that
+    // the workload should trust, keyed by the SPIFFE ID of the foreign trust
+    // domain. Bundles are ASN.1 DER encoded.
+    map<string, bytes> federated_bundles = 3;
+}
+
+// The X509SVID message carries a single SVID and all associated information,
+// including the X.509 bundle for the trust domain.
+message X509SVID {
```
strideynet, line 357: For `x509_svid`/`bundle` fields, I think we should be more explicit whether or not PEM-wrapping should be present. I can remember running into this when first implementing SPIFFE - since PEM-wrapping is so common that one may mistakenly expect it.

standards/SPIFFE_Broker_API.md
```diff
@@ -0,0 +1,642 @@
...
+
+The specification currently defines two builtin workload reference types.
+
+**Process ID (PID) Reference**: Identifies a workload by its process identifier. The PID MUST be a positive integer. This reference type is universally supported across POSIX-compliant systems.
+
+Example:
+```protobuf
+WorkloadReference {
+  reference: Any {
+    type_url: "type.googleapis.com/WorkloadPIDReference"
```
strideynet: Is this type URL correct? Surely this should be `type.googleapis.com/spiffe.broker.WorkloadPIDReference`.

standards/SPIFFE_Broker_API.md
strideynet, line 1: Nit: Inconsistent markdown wrapping across file.

strideynet (commented):
Mostly there - a few (mostly stylistic/verbiage) comments.

strideynet:
Procedurally - should we try to fast-track discussion on https://github.com/spiffe/spiffe/pull/399 so we can release SPIFFE Broker API as a "Beta" stability API?

standards/brokerapi.proto
matheuscscp: I avoided `kind` here because it has a specific meaning in Kubernetes. The kind does not fully qualify the Kubernetes object/resource type itself. Example:

For the type  `deployments.apps` the kind is `Deployment`. This is the Kubernetes "built-in" Deployment. The actions-runner-controller for example has the type `deployments.actions.summerwind.dev`. The kind is the same: `Deployment`.

standards/brokerapi.proto
matheuscscp: I think it's well-known enough among Kubernetes users and controller authors that `group` does not include the version. Users are very used to writing RoleBindings, for example, where only the RBAC API group is used: `rbac.authorization.k8s.io`

standards/brokerapi.proto
strideynet: Ah interesting - makes sense to me.

kfox1111:
> Procedurally - should we try to fast-track discussion on #399 so we can release SPIFFE Broker API as a "Beta" stability API?

I think we probably could provisionally release it before #399 is official, marking it as Beta anyway?

I wouldnt mind fast tracking #399 though.

standards/SPIFFE_Broker_API.md
sorindumitru: Yes, was discussed multiple times in in sig-spec and there was consensus on using Any

matheuscscp:
I have people watching this PR so I'm gonna give some updates discussed in the last few sig-spec meetings here.

There are two remaining blockers:

* Discussing #398. A decision will be made next week.
* SPIFFE is introducing experimental standards to allow for new standards to soak and have a brief moment where breaking changes are allowed. This standard will undergo an experimental period.

arndt-s:
@matheuscscp we shouldn't gate ourselves behind #398, and neither #399. If any of them are merged by then - great, but they don't need to go first by any means IMO.

standards/SPIFFE_Broker_API.md
arndt-s: Good catch, fixed in 943c913380a6178334fad7d3f92ddfb508a735fa.

standards/SPIFFE_Broker_API.md
arndt-s: Fixed in f992e424fd0890699cc809e2b4cf6aa824085669

standards/SPIFFE_Broker_API.md
arndt-s: Agreed. This is too implementation specific and shouldn't be part of the spec. Removed in d8b315c227ed7005fbf0bfd33d1affede93ff46a

standards/SPIFFE_Broker_API.md
arndt-s: Do you mean we should be more explicit that it is raw DER and not PEM encoded?

standards/SPIFFE_Broker_API.md
arndt-s: Good point. I rephrased section 2 a bit so that it's more clear what applies and also reference 3.1.4 - ae1981ad809e0c3d7204fe7df25aecf2545d390a

matheuscscp:
> @matheuscscp we shouldn't gate ourselves behind #398, and neither #399. If any of them are merged by then - great, but they don't need to go first by any means IMO.

Awesome in my opinion, but I think that's what was said in the last sig-spec meeting :cry: Unless I got it wrong @strideynet?

standards/SPIFFE_Broker_API.md
strideynet: Yep

matheuscscp:
@arndt-s About #398, it was said in the meeting that this week's meeting would effectively unblock from it with a decision if the proposer does not further pursue the proposal, so it's not a major blocker (tried to convey this in the update above but apparently I failed)

matheuscscp:
And yeah according to #399 there are no blockers, if this PR gets merged without expressing the stability level then it won't qualify as Stable :+1:

standards/SPIFFE_Broker_API.md
sorindumitru: Thanks, looks good to me now

matheuscscp:
Not sure but it's possible that to resolve this thread in the SPIRE PR we may need a change here:

https://github.com/spiffe/spire/pull/6915#discussion_r3369171255

standards/brokerapi.proto
arndt-s: Discussed today amongst the maintainers.

We have a native Kubernetes reference now and don't think we need to go into the details of pause container PID references here. If anything it's an implementation details whether it is supported or not.

standards/SPIFFE_Broker_API.md
arndt-s: Discussed offline, outdated.

standards/SPIFFE_Broker_API.md
arndt-s: Fixed in 72fef297472254f007a5992ced4980ca85fa43cf.

standards/SPIFFE_Broker_Endpoint.md
```diff
@@ -0,0 +1,113 @@
...
+
+As a hardening measure against [Server Side Request Forgery](https://www.owasp.org/index.php/Server_Side_Request_Forgery) (SSRF) attacks, every client request to the SPIFFE Broker Endpoint MUST include the static gRPC metadata key `broker.spiffe.io` with a value of `true` (case sensitive). Requests not including this metadata key/value MUST be rejected by the SPIFFE Broker Endpoint (see the [Error Codes](#6-error-codes) section for more information). This prevents an attacker from exploiting an SSRF vulnerability to access the SPIFFE Broker Endpoint unless the vulnerability also gives the attacker control over outgoing gRPC metadata.
+
+### 3.1 Transport Security
+
+The SPIFFE Broker Endpoint requires transport security in the form of mutual TLS. Compared to the SPIFFE Workload Endpoint the broker and the implementation are expected to have a previously established root of trust in the form of SPIFFE bundles and X509-SVIDs. Effectively, this means that both parties need to make use of the SPIFFE Workload API as workloads first, before serving or connecting to the SPIFFE Broker Endpoint. Please see the [Authentication](#5-authentication) section for more information.
```
strideynet:
```suggestion
The SPIFFE Broker Endpoint requires transport security in the form of mutual TLS. Compared to the SPIFFE Workload Endpoint the broker and the implementation are expected to have a previously established root of trust in the form of SPIFFE bundles and X509-SVIDs. Effectively, this means that both parties need to make use of the SPIFFE Workload API as workloads first, before serving or connecting to the SPIFFE Broker Endpoint. Please see the [Authentication and Authorization](#5-authentication-and-authorization) section for more information.
```

standards/SPIFFE_Broker_API.md
```diff
@@ -0,0 +1,569 @@
...
+
+Implementations MUST NOT add or modify RPCs outside of this standard. Custom workload reference types MAY be defined using the extension point described in [Section 3.1.4](#314-extensibility). Any other extensions MUST be introduced as new gRPC services, as described in the [extensibility method](./SPIFFE_Broker_Endpoint.md#7-extensibility-and-services-rendered) of the SPIFFE Broker Endpoint specification.
+
+## 3. Service Definition
+
+The SPIFFE Broker API is defined by a Protocol Buffer (version 3) service definition. The complete definition is found in [brokerapi.proto](brokerapi.proto).
+
+Profiles are implemented as a group of related RPCs within a single `API` service in the `spiffe.broker` package.
+
+## 3.1 Workload Reference
```
strideynet:
```suggestion
### 3.1 Workload Reference
```

standards/SPIFFE_Broker_API.md
```diff
@@ -0,0 +1,569 @@
...
+
+Implementations MUST maintain a strict allow-only policy that prevents any caller that is not authorized to leverage the SPIFFE Broker API.
+
+Implementations MAY enforce more fine-grained access control by inspecting data carried in the request. For example, an implementation MAY restrict a given caller to a subset of reference types (such as only `WorkloadPIDReference`, or only `KubernetesObjectReference` against a specific namespace), to a subset of audiences in `FetchJWTSVID`, or to a subset of requested SPIFFE IDs. The exact policy model is implementation-defined.
+
+### 4.2 Remote procedure scope
+
+Every invocation of a remote procedure (RPC) at the SPIFFE Broker API, including its request and responses (potentially multiple), are in context of a concrete workload. Clients are expected to invoke RPCs for each workload they represent individually and isolate them for each other accordingly. An X509 Bundle response, for instance, is only valid for the workload the request has referenced and MUST NOT be applied, visible or in any other way impact other workloads. Same applies to all other RPCs in the scope of the SPIFFE Broker API.
```
strideynet:
```suggestion
Every invocation of a remote procedure (RPC) at the SPIFFE Broker API, including its request and responses (potentially multiple), are in context of a concrete workload. Clients are expected to invoke RPCs for each workload they represent individually and isolate them from each other accordingly. An X509 Bundle response, for instance, is only valid for the workload the request has referenced and MUST NOT be applied, visible or in any other way impact other workloads. Same applies to all other RPCs in the scope of the SPIFFE Broker API.
```

standards/SPIFFE_Broker_API.md
```diff
@@ -0,0 +1,569 @@
...
+}
+```
+
+### 6.2 Profile RPCs
+
+#### 6.2.1 FetchJWTSVID
+
+The `FetchJWTSVID` RPC allows a Broker to request one or more short-lived JWT-SVIDs with a specific audience for a workload.
+
+The `FetchJWTSVIDRequest` request message contains a mandatory workload reference. It also contains a mandatory `audience` field, which MUST contain the value to embed in the audience claim of the returned JWT-SVIDs. The `spiffe_id` field is optional, and is used to request a JWT-SVID for a specific SPIFFE ID. If unspecified, the server MUST return JWT-SVIDs for all identities authorized for the workload.
```
strideynet:
```suggestion
The `FetchJWTSVIDRequest` request message contains a mandatory workload reference. It also contains a mandatory `audience` field, which MUST contain the values to embed in the audience claim of the returned JWT-SVIDs. The `spiffe_id` field is optional, and is used to request a JWT-SVID for a specific SPIFFE ID. If unspecified, the server MUST return JWT-SVIDs for all identities authorized for the workload.
```

strideynet (dismissed):
LGTM - a few stylistic fixes.

Please ensure we include header indicating "in beta" or similar as discussed in interim call we had prior to merging.

standards/SPIFFE_Broker_API.md
```diff
@@ -0,0 +1,574 @@
...
+
+References MUST be resolved on the server and the server MUST verify the existence of the referenced workload.
+
+Servers MUST NOT trust reference data provided by the client without independent verification. For example, when a client provides a PID reference, the server SHOULD independently verify the process exists and collect workload identity attributes through secure channels (e.g., /proc filesystem, container runtime APIs) rather than accepting client-provided attributes at face value.
+
+### 3.1.2 Reference scope
+
+Some references — such as the process ID — are only meaningful and discoverable on the local node. Clients MUST ensure that local references are not sent to remote Broker API servers, and deployments MUST ensure servers do not receive requests originating from outside the node they run on. This mitigates situations where a Broker requests credentials with a process ID from a different node, where that process ID is used by a different workload.
```
sorindumitru, line 95: > Clients MUST ensure that local references are not sent to remote Broker API servers, and deployments MUST ensure servers do not receive requests originating from outside the node they run on.

What do we mean here by deployments? broker deployments? Maybe this should just say `servers MUST ensure they do not ...`

Also should this be `do not receive requests with local references from outside the node they run on`? Something like:
```suggestion
Some references — such as the process ID — are only meaningful and discoverable on the local node. Clients MUST ensure that local references are not sent to remote Broker API servers, and servers MUST ensure they do not process requests with local references originating from outside the node they run on. This mitigates situations where a Broker requests credentials with a process ID from a different node, where that process ID is used by a different workload.
```

standards/SPIFFE_Broker_API.md
```diff
@@ -0,0 +1,574 @@
...
+
+Object references (such as `KubernetesObjectReference`) are valid across the control plane that owns the referenced object — for example, anywhere within the same Kubernetes cluster — and MAY be sent across the network within that scope. Object references MUST NOT be honored by a server bound to a different control plane (e.g., a different cluster); servers MUST reject object references that name a control plane they do not serve.
```
sorindumitru: The `KubernetesObjectReference` does not have a field that names the control plane. Should it get an optional `cluster`?

standards/SPIFFE_Broker_API.md
```diff
@@ -0,0 +1,574 @@
...
+- `domain`: "spiffe.io"
+- `metadata`: Optional contextual information such as the PID value or other debugging details
+
+Clients MAY inspect ErrorInfo details for structured error information but MUST handle cases where only the status code is available.
+
+### 4.9 Workload Lifecycle
+
+Server and client MUST ensure that no operations are performed beyond the lifetime of the workload. Once the workload has stopped, the server MUST stop sending responses for that workload, and the client MUST drop all data it received for the workload, removing it from file systems or other locations where it may have been stored.
+
+What constitutes "stopped" depends on the reference type. For local references (such as a process ID), the workload is considered stopped when the underlying process terminates. For object references (such as a `KubernetesObjectReference`), the workload is considered stopped when the referenced object no longer exists in the control plane, or — when the reference pinned a UID alongside a name — when the object resolved by name no longer matches the pinned UID.
```
sorindumitru: Note for SPIRE (and possibly other implementations), this means that workloads may need to be re-attested at least periodically.

I don't know if we want to relax this requirement a bit, does this mean that we would have to watch these resources?

standards/SPIFFE_Broker_API.md
arndt-s: The reason I'm using "deployment" here is that it can also be the infrastructure that ensures it, e.g. by using a UNIX socket, or firewall policies. It doesn't have to be the server itself.

standards/SPIFFE_Broker_API.md
arndt-s: My understanding is that the server itself is bound to a dedicated Kubernetes control plane and that the relationship is always 1-1. Maybe that's too much inferring existing SPIFFE architectures but I'm ensure if there's value to open up the Broker API being able to issue credentials from multiple Kubernetes clusters.

standards/SPIFFE_Broker_API.md
arndt-s: I think that relaxing the MUST to a SHOULD would decrease security by quite a bit. How do you feel if we add some classifier like `in a timely manner`? It would give SPIRE and other implementation some leeway to prioritize scale & resiliency and not re-attest at every request?

standards/SPIFFE_Broker_API.md
sorindumitru: Yes, but here we are asking the server to not honour those requests that are not within its control plane, but offer no way for it to differentiate. I agree that it's unlikely to be an issue in practice, but having something in the reference, which in the case of k8s would likely be the cluster, would help.

Happy to also defer this for a later discussions, since I don't think it's that big of an impediment, and I don't think it's worth delaying the spec for it.

standards/SPIFFE_Broker_API.md
sorindumitru: That sounds good to me. I think the client (i.e. broker) should try to respond to this quickly and stop the stream, but the server can also do its own check.

For example in SPIRE I think a natural way of doing this is re-triggering attestation every time an updated SVID would be sent to the workload. I don't think we need to have that in the current PR for it, we can see later how we want to handle it.

strideynet: > How do you feel if we add some classifier like in a timely manner?

That makes sense to me.

standards/SPIFFE_Broker_API.md
arndt-s: I don't think we can directly defer because, as you're pointing out, there's no name

> servers MUST reject object references that name a control plane they do not serve.

standards/SPIFFE_Broker_API.md
arndt-s: 4a1882bbbb88e978703cdb902c53125b4767b017

standards/SPIFFE_Broker_API.md
```diff
@@ -0,0 +1,574 @@
...
+## 4. Client and Server Behavior
+
+### 4.1 Authenticating and authorizing the Caller
+
+The SPIFFE Broker API makes use of the authentication and authorization at the [SPIFFE Broker Endpoint](./SPIFFE_Broker_Endpoint.md#5-authentication-and-authorization) to identify and authorize the caller.
+
+Implementations MUST maintain a strict allow-only policy that prevents any caller that is not authorized to leverage the SPIFFE Broker API.
+
+Implementations MAY enforce more fine-grained access control by inspecting data carried in the request. For example, an implementation MAY restrict a given caller to a subset of reference types (such as only `WorkloadPIDReference`, or only `KubernetesObjectReference` against a specific namespace), to a subset of audiences in `FetchJWTSVID`, or to a subset of requested SPIFFE IDs. The exact policy model is implementation-defined.
+
```
arndt-s, line 227: @sorindumitru 
```suggestion

Because the [SPIFFE Broker Endpoint](./SPIFFE_Broker_Endpoint.md#3-transport) MAY be served over a TCP listen socket, the caller is not necessarily co-located with the implementation. Authorized brokers MAY therefore request SVIDs from outside the cluster or host on which the implementation runs, provided they satisfy the authentication and authorization requirements of the SPIFFE Broker Endpoint. Where this is not desired, an implementation can restrict such access either by differentiating on the SPIFFE ID of the broker as part of its authorization policy, or by limiting the endpoint to local accessibility, for instance by serving it over a Unix Domain Socket.

```

standards/SPIFFE_Broker_API.md
arndt-s: Fixed with d0d62c19cc92458d12bee38d6bd9557fc288fdea

strideynet (approved):

sorindumitru (approved):

matheuscscp:
:rocket: !

PR #342, opened 2025-06-04
Correct reference in X509-SVID.

jhoyla:
The current reference text reads `RFC 5280, section 4.2.1.12`, but the link points to `https://tools.ietf.org/html/rfc5280#section-4.2.1.2`. The correct section is 4.2.1.12, Extended Key Usage.


kfox1111 (approved):
LGTM. Thanks! :)

arndt-s (approved):
Great catch! Thank you!

PR #361, opened 2025-11-10
Introduce WIT-SVID Token Document

strideynet:
As discussed in SIG-Spec, this document will mostly refer to documents produced by the IETF WIMSE working group. We may take the opportunity to profile this further in a few cases.

Will merge into https://github.com/spiffe/spiffe/pull/362 - where we will have a final opportunity to review the drafts as a whole.

PR #362, opened 2025-11-14
Integrating WIT to the SPIFFE standards

### LINKED ISSUE #315: Support for WIMSE WITs

strideynet:
Introduces a series of changes to the SPIFFE specification to formalise the IETF WIMSE Workload Identity Token as the third SPIFFE SVID (WIT SVID).

Please see https://datatracker.ietf.org/wg/wimse/about/ for the upstream work at IETF is being implemented here.

This is a PR is compromised of several PRs that have already undergone review in their own right:

- https://github.com/spiffe/spiffe/pull/327
- https://github.com/spiffe/spiffe/pull/361

A final review pass on this PR will be completed to ensure that content between the documents is consistent and that there is nothing missing. We'll likely hold off merging this content until we have a working implementation and can verify that the specification changes are fit for purpose.

Rather than directly leaving comments on this PR, consider raising feedback within GitHub issues or changes against this branch with a PR. This will avoid this PR becoming unmanagable.

Closes https://github.com/spiffe/spiffe/issues/315


strideynet:
Marking this as "Ready for Review", as we're starting to see implementation work towards WIT-SVID and the "Draft" status is giving an unintended signal that this spec is less stable than it is.

No strong plans to merge this in the short/medium term yet - so reviewers may hold off.

arndt-s (approved):
🚀

sorindumitru (approved):

Issue #364, opened 2025-11-20
Document differences and considerations when selecting an SVID type


jsnctl:
As discussed as SIG-Spec 20/11/25 + with discussions on SPIFFE slack from @strideynet @yoctozepto @arndt-s https://spiffe.slack.com/archives/C7AMR9NKB/p1763650116315739

With the advent of the WIT-SVID, there's now three options for an SVID - each with their own features and contextual considerations for use.

Implementers and end-users would benefit from some documentation on the differences between each, providing guidance and clarification on the most suitable selection. 

This has been mooted as potentially:
* An appendix to part of the SPIFFE standard
* A separate document
* ...?

yoctozepto:
> This has been mooted as potentially:
> 
> * An appendix to part of the SPIFFE standard

This approach has been suggested by @strideynet for the further explanation of differences between WIT-SVID and JWT-SVID to be potentially included in the WIT-SVID document's appendix. I concur it makes sense to reiterate on the differences.

(shameful plug; we wrote something along the lines with @jsnctl and other colleagues at Cofide: https://www.cofide.io/resources/solving-the-middlebox-dilemma-with-wimse-and-http-message-signatures/)

> A separate document

Suggested to be similar to IETF's Informational/BCP (Best Current Practices). Possibly to be placed in spiffe/spiffe. Placement to be decided. Let's write it first.

The longer document needs to treat the 3 kinds of SVID.

PR #366, opened 2025-11-30
Draft: Best current practices - Selecting an SVID format

jsnctl:
implements https://github.com/spiffe/spiffe/issues/364

As promised at the last SIG-Spec, here's a initial draft for us to work with regarding a best practices document for selecting an SVID. 

I've kept things fairly light on content so far, so feel free to suggest as many additions or refactoring of the structure as you like. I figure having an initial skeleton to work with is a good place for us to begin iterating on the document together

Should also note this adds the `best-current-practices/` directory to the root of the project


best-current-practices/Selecting_an_SVID_format.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Status of this Memo
+This document provides information and guidance on the selection of [SPIFFE Verifiable Identity Document](/standards/SPIFFE.md#3-the-spiffe-verifiable-identity-document) (SVID) formats for use across common workload identity scenarios. Distribution of this document is unlimited.
+
+## Abstract
+The SPIFFE standard specifies multiple compliant SVID formats. Selection of the most appropriate for use is dependent on numerous factors, including but not limited to: client library availability \& maturity, desired security properties, interoperability, certificate vs. token semantics, and the nature of expected workload-to-workload networking.
+
+This document outlines the best current practices to consider when selecting an SVID format for use in a SPIFFE-based identity system.
+
+## Table of Contents
```
strideynet, line 11: Perhaps it'd be useful to have a section talking about using multiple SVID types within a trust domain - I think often people think this is uncommon or not the "right" thing to do and assessing the pros/cons here may give people more confidence in considering it.

best-current-practices/Selecting_an_SVID_format.md
```diff
@@ -0,0 +1,112 @@
...
+
+## 2. Overview of Formats
+This section provides a brief description of the available SVID formats for contextual and comparative clarity. Full specification details for each can be found in their corresponding standardization (linked) documents.
+
+### 2.1 X.509-SVID
+[X.509-SVIDs](/standards/X509-SVID.md) are short-lived X.509 certificates containing a SPIFFE ID in the Subject Alternative Name. They are issued by the certificate authority (CA) of their SPIFFE trust domain, and used to authenticate during mutual TLS (mTLS) between workloads.
+
+Casting SVID information to an X.509 certificate format also allows for transport-level encryption to be used between peers with mutual TLS support, meaning X.509-SVIDs enable secure communication in transit between workloads that have such capabilities.
+
+### 2.2 JWT-SVID
```
strideynet, line 47: I think mentioning the OIDC ID token compat here is valuable since this allows them to be used with a lot of third party platforms as a form of authentication

strideynet:
Thanks for raising this! I've done a first pass. One thing that has occurred to me is that we may need to update CODEOWNERS etc to cover the new directory.

best-current-practices/Selecting_an_SVID_format.md
jsnctl: A very good point, yeah I think discussion of this would be very useful

best-current-practices/Selecting_an_SVID_format.md
```diff
@@ -0,0 +1,112 @@
+# Best Current Practice - Selecting an SVID format
+
+## Status of this Memo
+This document provides information and guidance on the selection of [SPIFFE Verifiable Identity Document](/standards/SPIFFE.md#3-the-spiffe-verifiable-identity-document) (SVID) formats for use across common workload identity scenarios. Distribution of this document is unlimited.
+
+## Abstract
+The SPIFFE standard specifies multiple compliant SVID formats. Selection of the most appropriate for use is dependent on numerous factors, including but not limited to: client library availability \& maturity, desired security properties, interoperability, certificate vs. token semantics, and the nature of expected workload-to-workload networking.
```
arndt-s: I feel like this should be ordered differently. IMO:
- Desired security promises
- The nature of expected workload-to-workload networking (I'd call it infrastructure)
- Interoperability
- SDK availability & maturity (not just client)
- Certificate & token semantics

R/ "Certificate & token semantics" - do you have an example where this matters? Is this comparability with  OIDC & standard private PKI?

best-current-practices/Selecting_an_SVID_format.md
```diff
@@ -0,0 +1,112 @@
...
+
+More recently, WIT-SVID has been proposed as an additional token-based SVID format to make use of authentication concepts developed as part of the IETF [Workloads In Multi-System Environments (WIMSE)](https://datatracker.ietf.org/group/wimse/documents/) working group. This represents an enhancement over the existing JWT-SVID semantics by codifying proof-of-possession information as part of the mandatory claims of the token.
+
+## 2. Overview of Formats
+This section provides a brief description of the available SVID formats for contextual and comparative clarity. Full specification details for each can be found in their corresponding standardization (linked) documents.
+
+### 2.1 X.509-SVID
+[X.509-SVIDs](/standards/X509-SVID.md) are short-lived X.509 certificates containing a SPIFFE ID in the Subject Alternative Name. They are issued by the certificate authority (CA) of their SPIFFE trust domain, and used to authenticate during mutual TLS (mTLS) between workloads.
```
arndt-s:
```suggestion
[X.509-SVIDs](/standards/X509-SVID.md) are short-lived X.509 certificates containing a SPIFFE ID in the Subject Alternative Name. They are issued by the certificate authority (CA) of their SPIFFE trust domain, and often used to authenticate client and/or server workloads with mutual TLS.
```

best-current-practices/Selecting_an_SVID_format.md
```diff
@@ -0,0 +1,112 @@
...
+2.3. [WIT-SVID](#23-wit-svid)  
+3\. [Comparison Matrix](#3-compirison-matrix)  
+4\. [Selection Criteria](#4-selection-criteria)  
+4.1 [When to use X.509-SVID](#41-when-to-use-x509-svid)  
+4.2 [When to use JWT-SVID](#42-when-to-use-jwt-svid)  
+4.3 [When to use WIT-SVID](#43-when-to-use-wit-svid)  
+4.4 [Heuristics](#44-heuristics)  
+5\. [Appendix](#5-appendix)  
+
+## 1. Background
```
arndt-s, line 26: I think there's value stating that SPIFFE itself does not prelude any usage and just defines the credential formats. The standard doesn't talk about using X.509 for mTLS or how a JWT-SVID is attached to requests.

I believe that one of the goals of this document is to bring that theory more into practice.

best-current-practices/Selecting_an_SVID_format.md
```diff
@@ -0,0 +1,112 @@
...
+## 2. Overview of Formats
+This section provides a brief description of the available SVID formats for contextual and comparative clarity. Full specification details for each can be found in their corresponding standardization (linked) documents.
+
+### 2.1 X.509-SVID
+[X.509-SVIDs](/standards/X509-SVID.md) are short-lived X.509 certificates containing a SPIFFE ID in the Subject Alternative Name. They are issued by the certificate authority (CA) of their SPIFFE trust domain, and used to authenticate during mutual TLS (mTLS) between workloads.
+
+Casting SVID information to an X.509 certificate format also allows for transport-level encryption to be used between peers with mutual TLS support, meaning X.509-SVIDs enable secure communication in transit between workloads that have such capabilities.
+
+### 2.2 JWT-SVID
+[JWT-SVIDs](/standards/JWT-SVID.md) are signed JSON Web Tokens (JWT) containing a SPIFFE ID in the `sub` claim. It is issued and signed by the JWT issuer of the SPIFFE trust domain, and is used to authenticate in the application layer between workloads.
```
arndt-s: "between workloads" gives the impression it's mutual auth. Never seen it in practice and if we want to go there we should mention that data is sent before server is authenticated. Obviously dependent on the protocol but most rely on a lower-level handshake.

best-current-practices/Selecting_an_SVID_format.md
```diff
@@ -0,0 +1,112 @@
...
+
+## 2. Overview of Formats
+This section provides a brief description of the available SVID formats for contextual and comparative clarity. Full specification details for each can be found in their corresponding standardization (linked) documents.
+
+### 2.1 X.509-SVID
+[X.509-SVIDs](/standards/X509-SVID.md) are short-lived X.509 certificates containing a SPIFFE ID in the Subject Alternative Name. They are issued by the certificate authority (CA) of their SPIFFE trust domain, and used to authenticate during mutual TLS (mTLS) between workloads.
+
+Casting SVID information to an X.509 certificate format also allows for transport-level encryption to be used between peers with mutual TLS support, meaning X.509-SVIDs enable secure communication in transit between workloads that have such capabilities.
```
arndt-s: Alongside encryption comes integrity protection that the JWT-SVID doesn't and the WIT-SVID maybe has.

best-current-practices/Selecting_an_SVID_format.md
```diff
@@ -0,0 +1,112 @@
...
+
+### 2.2 JWT-SVID
+[JWT-SVIDs](/standards/JWT-SVID.md) are signed JSON Web Tokens (JWT) containing a SPIFFE ID in the `sub` claim. It is issued and signed by the JWT issuer of the SPIFFE trust domain, and is used to authenticate in the application layer between workloads.
+
+JWT-SVIDs are bearer tokens, and do not enable any kind of encryption or cryptographic protection against impersonation. Various recommendations exist to mitigate against known vulnerabilities of the JWT as an identity token, and should be followed when JWT-SVIDs are used in SPIFFE.
+
+### 2.3 WIT-SVID
+WIT-SVIDs are signed JWTs containing a SPIFFE ID in the `sub` claim and a public key binding, with proof-of-possession metadata regarding the signing key available in a `cnf` claim.
+
+WIT-SVIDs differ from JWT-SVIDs in that they are proof-of-possession tokens (i.e. **not** bearer tokens). This provides enhanced security guarantees about caller provenance, integrity, and non-repudiation while still maintaining the interoperability of using token-based authentication in the application layer.
```
arndt-s: WIT-SVIDs don't necessarily come with integrity protection, only in combination with a PoP-mechanism that enables it and even there, not out of the box.

arndt-s: "while still maintaining the interoperability of using token-based authentication in the application layer."

Interoperability can mean many things here. Maybe "while being decoupled from the networking layer" if that is what you mean?

best-current-practices/Selecting_an_SVID_format.md
```diff
@@ -0,0 +1,112 @@
...
+| **Rotation**          | Frequent, automatic                            | Short-lived tokens                             | Short-lived tokens, key material                           |
+| **Verification**      | Certificate chain                              | JWS signature, claims                          | JWS signature, claims, proof-of-possession                 |
+| **Replay resistance** | Strong (TLS handshake, key possession)         | Weak (stateless token expiry)                  | Strong (key possession)                                    |
+| **Maturity**          | High                                           | High                                           | Emerging                                                   |
+
+## 4. Selection Criteria
+This section provides guidance on scenarios where a single SVID format can be considered favorable when compared with the alternatives. This should not be considered exhaustive, and there are likely to be cases where multiple or any of the formats could be argued to be suitable.
+
+### 4.1 When to use X.509-SVID
```
arndt-s, line 78: Do we want to talk about mixed-modes? SPIFFE-based client auth in combination with WebPKI-based server auth? Or standard TLS with SPIFFE-server auth?

best-current-practices/Selecting_an_SVID_format.md
```diff
@@ -0,0 +1,112 @@
...
+
+## 3. Comparison Matrix
+
+|                       | X.509-SVID                                     | JWT-SVID                                       | WIT-SVID                                                   |
+|-----------------------|------------------------------------------------|------------------------------------------------|------------------------------------------------------------|
+| **Primary use**       | Transport-level identity with mutual TLS       | Request/API-level identity with bearer tokens  | Request/API-level identity with proof-of-possession        |
+| **Layer**             | Transport                                      | Application                                    | Application                                                |
+| **Format**            | X.509 certificate                              | Signed JWT                                     | Signed JWT with public key binding                         |
+| **Proof semantics**   | Holder-of-key (private key bound to cert)      | Bearer token                                   | Holder-of-key (private key bound to workload)              |
```
arndt-s: "private key bound to cert" vs "private key bound to workload" is confusing. Both keys have the same characteristics IMO.

arndt-s: Proof semantics "Bearer token" is a WPT or DPoP token. JWT-SVIDs don't have any proof semantics.

best-current-practices/Selecting_an_SVID_format.md
```diff
@@ -0,0 +1,112 @@
...
+| **Rotation**          | Frequent, automatic                            | Short-lived tokens                             | Short-lived tokens, key material                           |
```
arndt-s: I am not sure that there's any difference in the rotation characteristics between the tokens. JWT-SVIDs and WIT-SVIDs rotations are automatic and frequent too. The key material of X.509 doesn't differ from the one in WIT?

best-current-practices/Selecting_an_SVID_format.md
```diff
@@ -0,0 +1,112 @@
...
+| **Verification**      | Certificate chain                              | JWS signature, claims                          | JWS signature, claims, proof-of-possession                 |
+| **Replay resistance** | Strong (TLS handshake, key possession)         | Weak (stateless token expiry)                  | Strong (key possession)                                    |
```
arndt-s: There's a difference between 
* X.509 with TLS which uses nonces during the handshake and is out-of-the-box replay protected
* JWT-SVID which is not replay protected at all unless one deploys something on top. The affected scope is often broad, based on the audience
* WIT-SVID which is not replay protected either. However, dependent on the PoP mechanism and scope of it the applicability of the replay is more narrow compared to the JWT-SVID

best-current-practices/Selecting_an_SVID_format.md
```diff
@@ -0,0 +1,112 @@
...
+JWT-SVIDs are bearer tokens, and do not enable any kind of encryption or cryptographic protection against impersonation. Various recommendations exist to mitigate against known vulnerabilities of the JWT as an identity token, and should be followed when JWT-SVIDs are used in SPIFFE.
+
+### 2.3 WIT-SVID
+WIT-SVIDs are signed JWTs containing a SPIFFE ID in the `sub` claim and a public key binding, with proof-of-possession metadata regarding the signing key available in a `cnf` claim.
+
+WIT-SVIDs differ from JWT-SVIDs in that they are proof-of-possession tokens (i.e. **not** bearer tokens). This provides enhanced security guarantees about caller provenance, integrity, and non-repudiation while still maintaining the interoperability of using token-based authentication in the application layer.
+
+## 3. Comparison Matrix
+
+|                       | X.509-SVID                                     | JWT-SVID                                       | WIT-SVID                                                   |
```
arndt-s: I think we should make it more clear in the header that it's "X.509 with mTLS", "JWT-SVID in request header" and "WIT-SVID with PoP". Many of the rows are specific to the presentation, not the credential format itself.

arndt-s: I think we should add the following rows:
* Integrity protection: X.509 yes, JWT no, WIT maybe
* Encryption: X.509 yes, JWT & WIT no
* Server authentication: X.509 yes, JWT & WIT no
* Support for TLS-termination between workloads: X.509 no, JWT & WIT yes

best-current-practices/Selecting_an_SVID_format.md
arndt-s: We should mention it but I believe there's value in having separate documents for:
* Interoperability with JWT
* Interoperability with OAuth/OIDC
* Interoperability with PKI

Each of these topics can fill a document and have their own caviats.

arndt-s (commented):
Section 1-3

best-current-practices/Selecting_an_SVID_format.md
jsnctl: Yeah I agree a separate interop document (ultimately reference from here) would be a good structure

best-current-practices/Selecting_an_SVID_format.md
jsnctl: Ordering wasn't intended to communicate importance, but I think it probably reads as such - so your ordering works better (plus amendments).

Do you think bullets are clearer?

Re: semantics - I was thinking something like the combination of security + interop  differences between certs and tokens (i.e. cert vs bearer token, ease of extensibility in token vs comparative rigidity of certs). This might be superfluous though, and maybe confuses things so would be happy to remove

best-current-practices/Selecting_an_SVID_format.md
jsnctl: Sounds good, do you think a sentence stating this is enough to add to this paragraph?

best-current-practices/Selecting_an_SVID_format.md
jsnctl: Not intended as such, just poorly worded. I'll amend

jsnctl: https://github.com/spiffe/spiffe/pull/366/commits/991a66309779e608473afee25949cc3b9a0b28b8

best-current-practices/Selecting_an_SVID_format.md
jsnctl: https://github.com/spiffe/spiffe/pull/366/commits/635d5141f506f248c1b66231422f41cf56a559af

best-current-practices/Selecting_an_SVID_format.md
jsnctl: Softened the language a bit around this to talk about it being enabled rather than a turn-key property https://github.com/spiffe/spiffe/pull/366/commits/92c7a800cbae8b2eea3a676226bec9deda389271

jsnctl: I was meaning the flexibility w.r.t legacy and third party services that require auth in conjunction with processes like token exchange, although we likely need to say that.

Decoupling from the networking layer is a nice descriptor here, I'll add that specifically for the moment

jsnctl: https://github.com/spiffe/spiffe/pull/366/commits/bd31da984d21e3cfba7471d312a096dab5adc3ac

best-current-practices/Selecting_an_SVID_format.md
jsnctl: Table headers adjusted https://github.com/spiffe/spiffe/pull/366/commits/85958d8c23a455222b00adbfe0e4f8c82c96c3a4

best-current-practices/Selecting_an_SVID_format.md
jsnctl: Reading this back I think this was a misconception on my part about token vs. certs here, I'll remove this row

best-current-practices/Selecting_an_SVID_format.md
jsnctl: Removed to just `holder-of-key" or N/A (for JWT-SVIDs)

best-current-practices/Selecting_an_SVID_format.md
jsnctl: Yes I think so, best placed here or a distinct section?

best-current-practices/Selecting_an_SVID_format.md
jsnctl: Changed ordering and removed final point for clarity https://github.com/spiffe/spiffe/pull/366/commits/587c457f1eb6d3f86e765e96205970e48badc81d https://github.com/spiffe/spiffe/pull/366/commits/68132d9454ab221a8d4aebe569d550f3f424d18b

best-current-practices/Selecting_an_SVID_format.md
jsnctl: Added this https://github.com/spiffe/spiffe/pull/366/commits/fdff96bc92f51795b63c0a2dac3a0e350670c877

best-current-practices/Selecting_an_SVID_format.md
jsnctl: Added with https://github.com/spiffe/spiffe/pull/366/commits/cddaffa67275c7c99f8ccbe5ba06c2475a06f4b8

best-current-practices/Selecting_an_SVID_format.md
```diff
@@ -0,0 +1,117 @@
...
+| **Verification**          | Certificate chain                              | JWS signature, claims                          | JWS signature, claims, proof-of-possession                 |
+| **Replay resistance**     | Strong (TLS handshake, key possession)         | Weak (stateless token expiry)                  | Strong (key possession)                                    |
+| **Integrity protection**  | Yes                                            | No                                             | Yes (via presentation methods)                             |
+| **Encryption**            | Yes                                            | No                                             | No                                                         |
+| **Server authentication** | Yes                                            | No                                             | No                                                         |
+| **TLS termination**       | No                                             | Yes                                            | Yes                                                        |
```
strideynet: I think we may need to come up with a different title than "TLS Termination" or include an explanation of what we mean by this below.

best-current-practices/Selecting_an_SVID_format.md
```diff
@@ -0,0 +1,117 @@
...
+### 4.2 When to use JWT-SVID
+JWT-SVIDs are the preferred choice for systems where application layer authentication is required (in contrast to authentication in the transport layer) without the need for strong proof-of-possession verification. This might apply in scenarios where:
+
+* Systems do not have native or platform-enabled TLS support across all workloads,
+* Identity metadata needs to be inspected in-flight to perform actions such as routing and proxying,
+* Environments are highly heterogeneous, and third party identity integration (e.g. OIDC) or legacy systems are common
+
+Layer 7 networking is the typical context requiring token-based authentication throughout each of the participating applications.
+
+### 4.3 When to use WIT-SVID
```
strideynet, line 94: I'd perhaps emphasise here that due to maturity/available tooling, WIT-SVID is best used in environments where you control the implementation of services.

jsnctl: Do you think that should be an amendment to this part in particular? 

https://github.com/jsnctl/spiffe/blob/fdff96bc92f51795b63c0a2dac3a0e350670c877/best-current-practices/Selecting_an_SVID_format.md?plain=1#L98

best-current-practices/Selecting_an_SVID_format.md
jsnctl: I've added a footnote on TLS termination here, we can tweak the wording to be more exacting if we like too https://github.com/spiffe/spiffe/pull/366/commits/1e0316da3b2333d62f28f00222d3adba7fd16a11

Issue #370, opened 2026-01-21
WIT-SVID Draft: Prohibit `aud` claim in WIT


strideynet:
For https://github.com/spiffe/spiffe/pull/362

Original discussion: https://github.com/spiffe/spiffe/pull/361#discussion_r2704973837

strideynet:
Closed by https://github.com/spiffe/spiffe/pull/372

PR #372, opened 2026-01-21
WIT-SVID Draft: Prohibit inclusion of `aud` claim

strideynet:
Closes https://github.com/spiffe/spiffe/issues/370

Since we expect the Proof of Possession mechanism to be used to scope to a specific recipient, there is no purpose to the `aud` claim being included within the WIT itself - and allowing it creates a scenario where validators would need to determine how to behave if `aud` within the WIT and the PoP disagreed.

### DIFF
--- a/standards/WIT-SVID.md
+++ b/standards/WIT-SVID.md
@@ -186,6 +186,8 @@ Within SPIFFE, there already exists mechanisms for the distribution of trust bun
 
 ### 3.8. Additional Claims
 
+The audience (`aud`) claim MUST NOT be included within a WIT-SVID. The inclusion of this claim within the WIT-SVID is not appropriate as scoping authentication to a specific recipient should be implemented within the Proof of Possession and not the WIT itself.
+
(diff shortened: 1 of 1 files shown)


standards/WIT-SVID.md
```diff
@@ -186,6 +186,8 @@ Within SPIFFE, there already exists mechanisms for the distribution of trust bun
 
 ### 3.8. Additional Claims
 
+The audience (`aud`) claim MUST NOT be included within a WIT-SVID. The inclusion of this claim within the WIT-SVID is not appropriate as scoping authentication to a specific recipient should be implemented within the Proof of Possession and not the WIT itself.
```
strideynet, line 189: Curious if this feels out of place within the "Additional Claims" section. Perhaps it makes more sense to have a heading for this claim above? lmk.

yoctozepto: I think it's fine. The other claims can be used - unlike this one.

yoctozepto (approved):

PR #374, opened 2026-02-06
Correct JSON Schema for JWT SVID

strideynet:
Thanks to @geeknik of Deep Fork Cyber for reporting this issue.

This PR updates two elements of the JSON schema.

For the `header` property, I have added `additionalProperties: false`. This is to reflect the following from the specification:

> Any header not described here, registered or private, MUST NOT be included in the JWT-SVID JOSE Header.

To the `aud` claim, I have added a `oneof` to allow the value to either be a string, or an array of strings (minimum 1 item). This better reflects the common structure of `aud` within JWTs and JWT-SVIDs, and the following from the specification:

> The `aud` claim MUST be present, containing one or more values.

I'm not convinced that this schema file is in common usage - or at least it is not leveraged within any of the core SDKs. It also doesn't seem overly correct with `header` and `payload` as two properties of the same object - so I'm not sure that this could be in use by anybody without some gymnastics. Perhaps we should review this in the near future?

### DIFF
--- a/standards/JWT-SVID.schema
+++ b/standards/JWT-SVID.schema
@@ -39,6 +39,7 @@ properties:
           - "JOSE"
     required:
       - "alg"
+    additionalProperties: false
   payload:
     type: "object"
     properties:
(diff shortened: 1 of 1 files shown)


yoctozepto:
> I'm not convinced that this schema file is in common usage - or at least it is not leveraged within any of the core SDKs. It also doesn't seem overly correct with `header` and `payload` as two properties of the same object - so I'm not sure that this could be in use by anybody without some gymnastics. Perhaps we should review this in the near future?

Two separate schema files would be formally needed, for the header and the payload - as they are two separate JSON documents from the point of view of JSON Schema. Obviously, one could build a common JSON document from the two, according to this schema.

yoctozepto (approved):

strideynet:
> Two separate schema files would be formally needed, for the header and the payload - as they are two separate JSON documents from the point of view of JSON Schema.

Thanks - that was my gut feeling but not too overly familiar with JSON Schema. I think this strategy would make the most sense to pursue in the near future since if we're going to publish a JSON schema, it should at least be usable 😅

PR #390, opened 2026-04-30
Nominate Sorin Dumitru

strideynet:
I hereby nominate Sorin Dumitru for the 2026H1 SSC election.

In the past couple of years, Sorin has been a very valuable and active participant in SIG-Spec. He brings a fresh perspective as an end-user of SPIFFE/SPIRE, but also a depth of knowledge from his position as a maintainer of SPIRE. I believe this perspective and depth of knowledge would be extremely valuable in SSC.


v0lkan (approved):
1,000,000% LGTM.

🌮 .

PR #399, opened 2026-05-29
Introduce Stability/Graduation process for SPIFFE specifications

strideynet:
With the recent work on Broker API and WIT-SVID profile, it's become apparent that we need some way of tracking and communicating the stability on specifications.

Previously, we roughly settled on content within the `main` branch being "Stable" and experimental or in-progress specs being held within branches/PRs. This has posed a number of challenges:

1. It reduces discoverability of newer specifications as they can only be found directly on GitHub within PRs.
2. It reduces confidence of implementors - it's unclear /when/ they should start to produce downstream implementations and what they should expect.
3. It poses some devex/logistical pain in managing the seperate branches.

This PR takes a stab at a stability/graduation process for SPIFFE specs. I've tried to keep it very light-weight as we don't need to burden ourselves with significant bureacracy. This process should let us move /faster/ rather than slow us down.

### DIFF
--- a/standards/STABILITY.md
+++ b/standards/STABILITY.md
@@ -4,7 +4,7 @@
 1\. [What this is](#1-what-this-is)  
 2\. [Stability levels](#2-stability-levels)  
 2.1. [Proposed](#21-proposed)  
-2.2. [Beta](#22-beta)  
+2.2. [Incubating](#22-incubating)  
 2.3. [Stable](#23-stable)  
 3\. [Marking stability](#3-marking-stability)  
 3.1. [Marking part of a specification](#31-marking-part-of-a-specification)  
@@ -25,10 +25,10 @@ implementations at any maturity, and vice versa.
   - *Breaking-change contract:* anything may change or be removed at any time, without notice.
   - *Where it lives:* an open PR or proposal. Not merged into the `main` branch.
   - *Audience:* discussion in SIG-Spec and proof of concept implementations.
-- **Beta**
+- **Incubating**
   - *Breaking-change contract:* breaking changes are avoided, but may be made in response to real-world implementation experience.
-  - *Where it lives:* merged into the `main` branch, marked Beta.
-  - *Audience:* implementers willing to track changes and provide feedback. Features relying on a Beta specification are typically gated behind a feature flag.
+  - *Where it lives:* merged into the `main` branch, marked Incubating.
+  - *Audience:* implementers willing to track changes and provide feedback. Features relying on an Incubating specification are typically gated behind a feature flag.
 - **Stable**
   - *Breaking-change contract:* breaking changes are severely avoided and reserved for resolving critical security issues.
   - *Where it lives:* merged into the `main` branch, marked Stable.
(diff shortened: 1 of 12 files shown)


standards/STABILITY.md
```diff
@@ -0,0 +1,104 @@
+# SPIFFE Specification Stability
```
strideynet, line 1: I've kept wording through-out fairly non-normative - happy to switch to normative language if prefered. I don't know if I quite consider this document itself a "spec".

standards/STABILITY.md
```diff
@@ -0,0 +1,104 @@
...
+back to Beta, but may be extended in a non-breaking way with new sections or
+profiles marked Beta.
+
+## 3. Marking stability
+
+A specification declares its level with a banner immediately beneath the title,
+above the existing "Status of this Memo" section. Use the variant matching the
+level:
+
+> **Stability: Beta** - see [STABILITY.md](STABILITY.md).
```
strideynet: Should we add more verbiage to these headers?

standards/STABILITY.md
```diff
@@ -0,0 +1,104 @@
...
+default. This makes it clear to operators that they are relying on functionality
+that may still change, and keeps a Beta feature from becoming a de-facto stable
+one purely by virtue of being widely deployed.
+
+### 2.3. Stable
+
+A specification is promoted from Beta to **Stable** by a pull request that
+updates its banner, approved by two maintainers. The criterion for promotion is
+that at least two independent, interoperable implementations exist and no
+further breaking changes are anticipated.
```
kfox1111: There is a slight issue here I think. I like the idea of the 2 implementation requirement, but

It may be possible for a spec to get implemented and become stable in spire, but no other implementations choose to implement it. If the feature is shipped long enough, what then?

I could be over analyzing it though, and we could consider dealing with it only when it happens.

strideynet: Definitely open to rewording.

As it stands, it's a little intentionally loose - one could imagine "go-spiffe" being able to consume the API and SPIRE implementing the server-side of an API as "two interoperable implementations".

It may make more sense to just say that "promotion is at the behest of SIG-Spec/maintainers, and that it is often a sign that a spec is ready for promotion when it has distinct, interoperable implementations" and effectively leave this fully in the hands of SIG-Spec to make the call.

kfox1111: Yeah. that wording is better I think.

standards/STABILITY.md
```diff
@@ -0,0 +1,104 @@
...
+A specification is promoted from Proposed to **Beta** by a pull request that
+merges it into the `main` branch with a Beta banner, approved by two
+maintainers. The criterion for promotion is SIG-Spec consensus that its design
+is sound and ready for trial implementation.
+
+We avoid breaking changes to Beta specifications. However, the purpose of the
+Beta phase is to gather experience from real implementations, and that
+experience may occasionally force a breaking change. Where it does, the change
+should be clearly communicated along with advice on adjusting implementations.
+Adopters should expect to track such changes and pin to a specific revision.
```
amartinezfayo: Since the specifications don't carry version numbers, I think it may be worth spelling out what pinning to a specific revision means in practice (presumably a git commit or permalink). It might also help to ask that breaking changes to a Beta specification be noted in the document itself, perhaps as a short changelog, so implementers can tell which revision they implemented without digging through git history.

standards/STABILITY.md
amartinezfayo, line 1: Would it make sense to also link this new STABILITY.md doc from the SPIFFE Standards list in the repo README? That section is likely the first place implementers land, and a pointer there would surface the stability levels without having to open each individual spec.

standards/STABILITY.md
```diff
@@ -0,0 +1,104 @@
...
+entirely.
+
+Because Proposed specifications live in PRs rather than in `main`, they require
+no in-document stability marking - the fact that a specification is unmerged
+indicates that it is Proposed.
+
+### 2.2. Beta
+
+A specification is promoted from Proposed to **Beta** by a pull request that
+merges it into the `main` branch with a Beta banner, approved by two
```
amartinezfayo: I'm wondering how this interacts with the SSC's role. GOVERNANCE.md describes [the SSC as exclusively responsible for SPIFFE's standards](https://github.com/spiffe/spiffe/blob/main/GOVERNANCE.md#the-spiffe-steering-committee-ssc), and the analogous level-change process in MATURITY.md requires SSC approval. Is the intent that the SSC delegates promotions to SIG-Spec and the maintainers of this repository? It might also help to spell out which maintainers "approved by two maintainers" refers to. A sentence making that relationship explicit would keep the two documents from appearing to disagree about who holds the authority.

amartinezfayo (commented):
Thank you @strideynet for putting this together. I gave it a read and have a few questions / suggestions.

standards/STABILITY.md
strideynet: It's a good question!

By "maintainers" here, I am referring to the list of individuals marked as code owners for the `/standards` directory: https://github.com/spiffe/spiffe/blob/main/CODEOWNERS#L5. I can update this to use the term "code owner" instead as that seems a little more explicit.

> GOVERNANCE.md describes [the SSC as exclusively responsible for SPIFFE's standards](https://github.com/spiffe/spiffe/blob/main/GOVERNANCE.md#the-spiffe-steering-committee-ssc),

I think there's definitely an action item here to revisit the governance documents and the role of SSC and SIG-Spec. The situation is definitely blurry at the moment with the seperate existence of the Standards code-owners/maintainers. I'll admit in the past this has been confusing to myself and @arndt-s as well. I would believe that historically - this would have been less of an issue as there would have been a greater degree of overlap between SSC and SIG-Spec/Standard code-owners.

My personal stance here is that it makes sense for SIG-Spec/SPIFFE standards code-owners to have some degree authority here. There's no guarantee that SSC members are technical or interested in standards work - and - I think that's a good thing. Gating on SSC

standards/STABILITY.md
strideynet: The inclusion of a "breaking changes" section, if we should ever make any, seems like a pretty sensible idea.

Agree on expanding on what I mean by pinning here.

standards/STABILITY.md
strideynet: 5898efc

standards/STABILITY.md
strideynet: cd9076a

standards/STABILITY.md
amartinezfayo: This all makes sense to me! I think the last sentence may have gotten cut off at "Gating on SSC"? I couldn't quite follow what it meant :)

strideynet: Ah - yes looks like I got halfway through writing that and got distracted. I think I was going to say something along the lines of "Gating on SSC may not make sense if a future SSC is more heavily balanced towards governance/community and less-so technically".

strideynet: [e26477a](https://github.com/spiffe/spiffe/pull/399/commits/e26477a2074a6029ff6963132257823767080571)

standards/STABILITY.md
strideynet: I've adjusted the wording in 309bb03 and bc8fe5f

kfox1111:
LGTM! :)

sorindumitru (dismissed):
LGTM, thanks @strideynet !

arndt-s (approved):
LGTM

sorindumitru (approved):

PR #401, opened 2026-06-12
Add clarification that keys & bundles are not PEM encoded

arndt-s:
Small editorial fixes to make it more clear that keys & bundles are NOT PEM but only ASN.1 DER encoded.


strideynet (approved):

sorindumitru (approved):
LGTM too

PR #404, opened 2026-06-16
Update SSC README with Sorin's membership

strideynet:
As per 26H1 election results.

### DIFF
--- a/ssc/README.md
+++ b/ssc/README.md
@@ -6,6 +6,10 @@ Please read the [SSC Charter](CHARTER.md) for more information about the SSC's r
 ## SSC Members
 The current members of the SSC, ordered by term,  are as follows:
 
+**Sorin Dumitru** (@sorindumitru)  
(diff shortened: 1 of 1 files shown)


ssc/README.md
```diff
@@ -6,6 +6,10 @@ Please read the [SSC Charter](CHARTER.md) for more information about the SSC's r
 ## SSC Members
 The current members of the SSC, ordered by term,  are as follows:
 
+**Sorin Dumitru** (@sorindumitru)  
+[GitHub](https://github.com/sorindumitru) | [LinkedIn](https://www.linkedin.com/in/sorin-dumitru-69841243)  
+Term end: June 16th, 2028
```
strideynet, line 11: TBD - We need to decide what date the term will officially begin.

strideynet: In absence of agreeing a date in the future, I think the 16th as the announcement day likely makes the most sense here. Next SSC election we'll be better about kicking off on the original timeline so this is less confusing.

v0lkan (approved):
🌮 🎊 .

PR #406, opened 2026-06-18
Document Private Enterprise Number

strideynet:
We have been allocated a Private Enterprise Number - I don't know if we'll ever use it - but it's useful to have this documented/around in case we ever wish to define some OIDs for X.509 extensions etc. It's also entirely possible that such work would make more sense under IETF umbrella/IETF OIDs.


sorindumitru (approved):
LGTM, thanks Noah!

PR #411, opened 2026-07-02
Relax KubernetesObjectReference uid validation

matheuscscp:
Relaxing "at the time of issuing the SVIDs" to "at the time of attesting the reference".

"For a KubernetesObjectReference, if both `uid` and `key` are specified, then the implementation MUST ensure that they match *at the time of attesting the reference*."

It's very difficult to make this check every time SVIDs are issued e.g. X509-SVIDs that are rotated and returned in the stream RPC.


matheuscscp:
@arndt-s @strideynet @sorindumitru PTAL :pray:

matheuscscp:
As discussed via Slack, this actually needs to be honored, it could be a major vulnerability otherwise. SPIRE has a plan for the implementation.

PR #415, opened 2026-07-24
Document SPIFFE_ENDPOINT env var

kfox1111:


standards/SPIFFE_Endpoint.md
```diff
@@ -0,0 +1,80 @@
...
+
+## 2. The SPIFFE_ENDPOINT Environment Variable
+
+Clients may be explicitly configured with an endpoint location, or may utilize the well-known environment variable `SPIFFE_ENDPOINT`. If not explicitly configured, conforming clients MUST fall back to the environment variable. See [Client Behavior and Precedence](#3-client-behavior-and-precedence) for the full precedence rules, including the deprecated `SPIFFE_ENDPOINT_SOCKET` fallback.
+
+### 2.1. Value Format
+
+The value of the `SPIFFE_ENDPOINT` environment variable is one or more [RFC 3986](https://www.ietf.org/rfc/rfc3986.txt) URIs. Multiple URIs MAY be provided as an ordered list separated by commas (`,`). Whitespace surrounding an individual URI is not significant; clients MUST trim leading and trailing whitespace from each list entry before interpreting it.
+
+Each URI's scheme identifies the delivery mechanism. When a list is provided, it expresses an ordered preference: clients MUST select and use the first entry whose scheme they support, and MUST ignore entries whose scheme they do not support. This allows a producer to offer several mechanisms and let each client choose the first one it understands.
```
strideynet, line 41: I think we should explicitly discuss what happens if a client supports a scheme but it does not "work" and multiple are provided. Does it try the next one it supports? Or is it a "It only ever uses the first one it supports".

kfox1111: Yeah. I was wondering what to do about it. There are benefits and drawbacks...

Simply always picking the first one supported is easiest to implement, and document.

If we optionally supported a different mechanism, then it would be easier to implement then a consistent model, but it would be much harder for the end user to reason about what software behaved in what way.

If it was required, then we need to work through all the details...

Does it load balance by randomizing the list ordering before going through it?
Does it pick one to attempt and on failure, restart from the list again or does it go one at a time from the list until all are tried, before resetting the list from the original and trying again
How long does it wait between attempts?
If it can talk to a unix socket but no svid is available, does it failover to the next eventually? if so, how long?

etc.

kfox1111:
With my admin hat on, one of the use cases I'd like to support is only having to specify things once if at all possible to cover most apps.

for example:

Say I install spiffefs using spire. This would support applications that only support fs delivery like prometheus exporters.
I also install spire-agent, and it supports the workload api.

Instead of having to craft specific environment variables for each application, having to know which options each application supports and preffers, I could just register a default env var. That would be super helpful as an admin:
```
SPIFFE_ENDPOINT=unix://run/spire/agent/sockets/main/public/api.sock,file://run/spiffefs
```

PR #417, opened 2026-07-31
Add wit-svid to the list of supported use values

kanywst:
WIT-SVID.md 6.1 requires the `use` parameter of a WIT signing key to be set to `wit-svid`, but 4.2.2 here still lists only `x509-svid` and `jwt-svid`. SPIFFE-ID.md 3.3 was updated to name all three SVID types in #362; this section was missed.


strideynet (approved):

arndt-s (approved):

strideynet:
Thank you for the contribution ❤️

PR #419, opened 2026-08-14
Allow using ML-DSA algs for JWT and WIT SVIDs

sorindumitru:


arndt-s (commented):
Looks good so far 👍🏼

standards/WIT-SVID.md
```diff
@@ -123,15 +126,18 @@ In addition to the requirements set out in RFC7800 and WIMSE Workload Credential
...
+RS512     | RSASSA-PKCS1-v1_5 using SHA-512
+ES256     | ECDSA using P-256 and SHA-256
+ES384     | ECDSA using P-384 and SHA-384
+ES512     | ECDSA using P-521 and SHA-512
+PS256     | RSASSA-PSS using SHA-256 and MGF1 with SHA-256
+PS384     | RSASSA-PSS using SHA-384 and MGF1 with SHA-384
+PS512     | RSASSA-PSS using SHA-512 and MGF1 with SHA-512
+ML-DSA-44 | ML-DSA-44 as specified in table 1 of [FIPS-204][9]
+ML-DSA-65 | ML-DSA-65 as specified in table 1 of [FIPS-204][9]
+ML-DSA-87 | ML-DSA-87 as specified in table 1 of [FIPS-204][9]
```
sorindumitru, line 140: We could also mention something about size here, but I think that's well known at this point.

standards/SPIFFE_Trust_Domain_and_Bundle.md
```diff
@@ -83,6 +84,9 @@ The `kty` parameter MUST be set, and its behavior follows [Section 4.1][6] of RF
 #### 4.2.2. Public Key Use
 The `use` parameter MUST be set. Its value indicates the type of identity document (or SVID) that it is authoritative for. At the time of this writing, three SVID types are supported: `x509-svid`, `jwt-svid`, and `wit-svid`. The values are case sensitive. Please see the respective SVID specifications for more information about `use` values. Clients encountering either a missing `use` parameter or an unknown `use` value MUST ignore the entire JWK element.
 
+#### 4.2.3. Algorithm
+The `alg` parameter MAY be set, and its behavior follows [Section 4.4][13] of RFC 7517. Some key types do not, on their own, identify the algorithm that a key is to be used with; where the key type in use requires the `alg` parameter in order to do so, it MUST be set. The respective SVID specification MAY place further requirements on its presence and value.
```
sorindumitru: We could also mention needing to validate that the signing algorithm of the SVID matches the `alg` field, but I think that's already specified by other various standard (X.509 and JWT [which also covers WIT]) so it's not necessary.

arndt-s: Considering that RFC 9964 says that
> The alg JSON Web Key (JWK) parameter or COSE Key Common parameter is REQUIRED for all AKP keys.

I think we should harden this here and say 

> The alg parameter's behavior follows [Section 4.4][13] of RFC 7517. It MUST be set for keys whose kty is AKP, as required by [Section 3][14] of RFC 9964, and for any other key type whose defining specification requires it; otherwise it MAY be set. The respective SVID specification MAY place further requirements on its presence and value.

arndt-s:
Another thought on this:

Let's say I deploy JWT & X.509 in my organization and want to move to PQ. While workload to workload authentication is covered some of my use cases are using the JWT SVID as assertion towards IDPs such as Entra, AWS, OAuth Authorization Servers. Some of them don't support it.

Is there something we can do from the standards side to support such diverse environments? Moving to PQ will most certainly not be a big-bang and the better we help implementations to support the transition the better.

PR #423, opened 2026-08-27
Add SPIFFE Discovery spec

arndt-s:
SPIFFE Federation profile & URI currently need to be provisioned manually and out of band. A server that is presented with an SVID of an unknown trust domain name is currently not able to verify it and process the request. This prevents the use of SPIFFE in environment where prior trust establishment is undesired or not possible. This is common in agentic use cases where remote invocations intent is built at runtime or in other scenarios on the open web such as crawling.

The SPIFFE Discovery specification targets to enable such use cases by making SPIFFE Federation configuration and other metadata discoverable. A goal is to have zero-configuration and allow peers to derive everything at runtime purely on a trust domain name.

### DIFF
--- a/standards/SPIFFE_Discovery.md
+++ b/standards/SPIFFE_Discovery.md
@@ -1,0 +1,118 @@
+# SPIFFE Discovery
+
+> **Stability: Proposed** - see [STABILITY.md](STABILITY.md).
+
+## Status of this Memo
+This document specifies an identity API standard for the internet community, and requests discussion and suggestions for improvements. Distribution of this document is unlimited.
+
+## 1. Background
+
+In an interconnected world workloads often need to communicate with other workloads and resources outside of their trust domain. The [SPIFFE Federation](./SPIFFE_Federation.md) specification is available to retrieve SPIFFE Trust Bundles from other domains by configuring federation out of band. However, manual configuration is not always feasible on the internet and sometimes trust needs to be established at runtime.
+
+## 2. Introduction
+
+The SPIFFE Discovery specification defines a well-known endpoint and configuration document that consumers can use to discover SPIFFE Trust Domain metadata. Properties include information such as the SPIFFE Federation Endpoint & profile, signing algorithms in use and more. Participants can use this document to learn about the trust domain and verify SVIDs issued by it.
+
+The well-known endpoint does not need to be configured out of band and can be computed from SPIFFE-IDs and SPIFFE Trust Domain Names.
+
+## 2.1 Relationship to WIMSE 
+
+TODO
+
+## 3. Targeted Use Cases
+
+TODO
+
+## 4. Requirements
+
+This document assumes that the SPIFFE Trust Domain Name is a fully-qualified domain name which is available in DNS and has a trusted TLS certificate available. The trust model that is associated with DNS and the web public key infrastructure becomes the base for discovery and with that the SPIFFE Trust Bundle too.
+
+## 5. SPIFFE Trust Domain Discovery Endpoint
+
+This section defines the location of the SPIFFE Trust Domain Discovery Endpoint and the requirements that apply to servers serving it and to clients consuming it.
+
+### 5.1. Endpoint URL
+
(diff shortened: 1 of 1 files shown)


standards/SPIFFE_Discovery.md
```diff
@@ -0,0 +1,118 @@
...
+Clients issue an HTTP `GET` for the [computed endpoint URL](#51-endpoint-url). Only a `200 OK` response carries a Configuration Document; any other status code MUST be treated as a failure to discover the trust domain. Clients MUST parse the body as a JSON object and MUST verify that `spiffe_trust_domain_name` equals the authority of the endpoint URL, as described in [6.1.1](#611-spiffe-trust-domain-name-spiffe_trust_domain_name).
+
+Clients SHOULD honor the server's caching directives, SHOULD refresh the document periodically, and SHOULD retry failed retrievals with an exponential backoff.
+
+TODOs:
+- security considerations for caching
+
+# 6. SPIFFE Trust Domain Configuration Document
+
+The SPIFFE Trust Domain Configuration Document is a JSON document that is served at the SPIFFE Trust Domain Discovery Endpoint. It carries non-sensitive metadata of the SPIFFE Trust Domain.
```
strideynet, line 66: IMO - setting some kind of direction on the "purposes" of this document may be useful. Is it prime intent serve information required by validators intending to consume SVIDs from a trust domain?

standards/SPIFFE_Discovery.md
```diff
@@ -0,0 +1,118 @@
...
+
+TODO
+
+## 3. Targeted Use Cases
+
+TODO
+
+## 4. Requirements
+
+This document assumes that the SPIFFE Trust Domain Name is a fully-qualified domain name which is available in DNS and has a trusted TLS certificate available. The trust model that is associated with DNS and the web public key infrastructure becomes the base for discovery and with that the SPIFFE Trust Bundle too.
```
strideynet, line 28: > has a trusted TLS certificate available

Do we expect in most cases this will be Web PKI? Or are we intending to support the discovery endpoint being protected by some locally trusted CA-issued certificate as well?

standards/SPIFFE_Discovery.md
```diff
@@ -0,0 +1,118 @@
...
+### 6.1.1 SPIFFE Trust Domain Name (`spiffe_trust_domain_name`)
+
+The `spiffe_trust_domain_name` property carries the SPIFFE Trust Domain Name. When served on the SPIFFE Trust Domain Discovery Endpoint it must be equal to the authority of the URI. Clients that observe a different SPIFFE Trust Domain Name MUST reject the document and MUST ensure that documents are not processed under the context of a different SPIFFE Trust Domain Name.
+
+### 6.1.2 SPIFFE Bundle Endpoint URI (`spiffe_bundle_endpoint_uri`)
+
+The `spiffe_bundle_endpoint_uri` property contains a URI pointing to the SPIFFE Bundle Endpoint as defined in [SPIFFE Federation](./SPIFFE_Federation.md). The URI MUST use the `https` scheme.
+
+TODOs:
+- Different domain?
```
strideynet, line 91: My gut feel would be to allow this to be located anywhere. This should simplify the re-use of existing bundle endpoints.

strideynet:
Overall keen on the direction / problem that this solves.

standards/SPIFFE_Discovery.md
arndt-s: I think most cases will be WebPKI. Locally-trusted CAs are most common in enterprises where I don't see the "out-of-band" trust establishment a big problem.

But I'm thinking that it probably doesn't matter as long as the client that invokes the endpoint and fetches the document trust the certificate? This will also cover environments where outbound proxies break TLS and serve it internally with a private CA cert.

sorindumitru, line 28: This is at odds with the [SPIFFE-ID spec](https://github.com/spiffe/spiffe/blob/main/standards/SPIFFE-ID.md#21-trust-domain) which says that the trust domain name is not a DNS name.

To avoid any incidents for existing SPIFFE users, we should also specify that using the discovery endpoint must be opt-in, at least initially, so SDKS, like `go-spiffe`, wouldn't use it unprompted.

